Issue #34, opened 2015-01-24
Drop event is not triggered when item is removed from DOM after dragging starts and dropped outside of browser window in Chrome


hakanderyal:
Thanks for awesome lib.

On Chrome, if I remove the item from DOM after dragging starts, and drop the item outside of the browser window, drop event is not triggered. (I'm checking it on DragDropStore).

No problems on FF, when mouse is moved out of the browser window, drop event triggers immediately.

As a workaround, I've removed the [isFirefox()](https://github.com/gaearon/react-dnd/blob/d309adcae85e07a212e9c941ef052127bbcd4719/modules/utils/NativeDragDropSupport.js#L79) check from the `_lastDragSourceCheckTimeout` in  [NativeDragDropSupport.js](https://github.com/gaearon/react-dnd/blob/d309adcae85e07a212e9c941ef052127bbcd4719/modules/utils/NativeDragDropSupport.js), but haven't tested much yet if it breaks normal functionality in Chrome.

Looked around the source code a little bit for another fix, but couldn't come up with any.

gaearon:
Thanks for reporting. I'll take a look on Monday.

gaearon:
Would you mind sending a PR of the code that works for you?

gaearon:
Released in 0.6.3, can you give it a go?

gaearon:
Btw thanks for following through on this.
Browsers are nasty.

hakanderyal:
I'm already using it, no problems so far.

Thank you for awesome work.

PR #36, opened 2015-01-27
Remove `isFirefox()` check from `_lastDragSourceCheckTimeout`

hakanderyal:
To fix Chrome not triggering a global `drop` if source node was removed and the dragged item s dropped outside of browser window.

### DIFF
--- a/modules/utils/NativeDragDropSupport.js
+++ b/modules/utils/NativeDragDropSupport.js
@@ -76,19 +76,19 @@ if (typeof window !== 'undefined') {
     if (isWebkit() && checkIfCurrentDragTargetRectChanged()) {
       // Prevent animating to incorrect position
       e.preventDefault();
-    } else if (isFirefox()) {
-
-      // Firefox won't trigger a global `drop` if source node was removed.
-      // It won't trigger `mouseup` either. It *will* however trigger `dragover`
-      // continually during drag, so our strategy is to simply wait until `dragover`
-      // has stopped firing.
-
-      clearTimeout(_lastDragSourceCheckTimeout);
-      _lastDragSourceCheckTimeout = setTimeout(
-        triggerDragEndIfDragSourceWasRemovedFromDOM,
-        140 // 70 seems enough on OS X with FF33, double it to be sure
-      );
     }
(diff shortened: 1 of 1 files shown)


hakanderyal:
Doesn't seem to break anything on Chrome. I haven't tested it for Safari/IE.

hakanderyal:
Fixes #34.

gaearon:
Thank you.

One problem with this approach is that once an item leaves the current window, it is no longer recognized as valid, although the user may bring it back without dropping. I think it's a valid use case because user might drag item out of window bounds by mistake.

This isn't really a new issue, but I only noticed it now that it appears in Chrome as well. Any ideas for an alternative workaround? We could probably listen to `mouseover` on document as well and compare last `mouseover` with last `dragover` time but this seems fragile.

hakanderyal:
You are right about moving out by mistake. Also, even we are triggering `dragend`, mouse pointer still shows the item as dragging.

Increasing the timeout value to take moving out by mistakes into account may help, but it's also fragile.

I'll think about it more.

gaearon:
What do you think about this approach?

https://github.com/gaearon/react-dnd/pull/37

hakanderyal:
It covers the 'move out by mistake', but listening the mouse events on window all the time might be too intrusive.

Will adding listeners when drag starts, and removing them when drag ends work?

I'll dive into the native drag&drop api to see if there is any other solution.

gaearon:
>Will adding listeners when drag starts, and removing them when drag ends work?

We can probably bind [here](https://github.com/gaearon/react-dnd/blob/another-source-dom-removal-fix/modules/utils/NativeDragDropSupport.js#L113) and unbind [here](https://github.com/gaearon/react-dnd/blob/another-source-dom-removal-fix/modules/utils/NativeDragDropSupport.js#L121).

If this approach works for you, can you please take my branch in #37 and add (un)subscription logic?
I'm a bit busy with another project right now..

I gave you access to the repo.

hakanderyal:
Done. Tested it with FF and Chrome, couldn't find any problems. I'm also testing it on production, no problems so far.

gaearon:
Superseded by #37.

PR #37, opened 2015-01-27
When drag source is gone from DOM, dropping outside the window shouldn't freeze interface

### LINKED ISSUE #34: Drop event is not triggered when item is removed from DOM after dragging starts and dropped outside of browser window in Chrome

gaearon:
This is another attempt to fix #34.

### DIFF
--- a/modules/utils/NativeDragDropSupport.js
+++ b/modules/utils/NativeDragDropSupport.js
@@ -98,15 +98,6 @@ if (typeof window !== 'undefined') {
 
     triggerDragEndIfDragSourceWasRemovedFromDOM();
   });
-
-  // Mouse events tell us that dragging has ended but `dragend` didn't dire.
-  // This may happen if source DOM was removed while dragging.
-
-  window.addEventListener('mousein', triggerDragEndIfDragSourceWasRemovedFromDOM);
-  window.addEventListener('mouseout', triggerDragEndIfDragSourceWasRemovedFromDOM);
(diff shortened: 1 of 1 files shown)


hakanderyal:
I've removed the listeners other than mousemove, but forgot about the scenario that source is removed DOM, and dropped in browser window. In this case, drop event is not triggered till the mouse is moved.

`mousein` together with `mouseover` seems to do the trick.

Still, there is a small but noticeable delay between item is dropped, and drop event is fired, on FF.

gaearon:
Also worth testing on IE.
It may have the same issues but it doesn't have `mousein` AFAIK (it has `mouseenter`).

I'll take a look later today.

As for the delay, I guess we'll have to live with it! Do you see any other options?

Issue #38, opened 2015-01-28
endDrag function on component not being called when item is removed from DOM


hakanderyal:
I'm using react-dnd for a kanban board app (trello like), and when moving a card to another list (thus unmounting source component during drag), the `endDrag` function is not called, as the listeners for `DragDropStore` are removed on unmount.

As a workaround, I've added a `_lastDraggedItem` and `getLastDraggedItem` to `DragDropStore`, and subscribed to `DragDropStore` on parent component, handling the endDrag there. As we clear the variables in `DragDropStore` before firing the `DRAG_END` event, I couldn't find any other way to get dragged item.

The reason I'm not using `DROP` event is, it's not fired when the item is dropped outside of browser window, only `DRAG_END` is fired.

It works but it's messy, breaking the `beginDrag` & `endDrag` symmetry.

Are there any other ways to trigger a function on `endDrag`, with access to dragged item? 

Or if is this a valid usecase for others, should we add my workaround to the repo?

gaearon:
I'll need some time to think on that, will look later in the evening.

gaearon:
Pardon for dragging my feet on this. I'm a bit busy with work stuff at the moment.
Is this urgent?

hakanderyal:
No problems, not urgent at all. I'm already the hacky way I mentioned, it works for my usecase.

gaearon:
OK, let me know if you have other ideas for fixing this as well.
I'll try to find some time next week to look into this!

hakanderyal:
Here is the a brain dump:

I see two different things here.

1- When source component is unmounted during drag, the `endDrag` defined on source component becomes irrelevant. Removing the source component may not be a common case, but for trello like functionality, I couldn't come up with any other solutions (Moving an item from one container to other). I thought about not removing, but hiding the source item while dragging, but couldn't find a reliable way of doing it.

While the action looks like moving an item, I'm essentially removing the item in one container, and adding a new one with same properties in another container. I also thought about if I can somehow carry the source item to another component without unmounting it, but haven't tried it, and don't know if it's even possible or feasible.

2- Assuming unmounting the source component during drag is the only way, I need to run a function when drag finishes. I'm achieving that by subscribing the `DragDropStore` of react-dnd in the parent component, and reacting to changes there. There isn't any way to get dragged item from the `DRAG_END` event, thus I added the `lastDraggedItem` stuff. 

[...]

PR #39, opened 2015-01-28
add _lastDraggedItem and _lastDraggedItemType to DragDropStore

hakanderyal:
example for #38.


gaearon:
I'm closing this, since it's a workaround for the real problem we'll be solving with #54 and later #38 itself.

PR #42, opened 2015-01-30
Add bower support

jgable:
Adds a bower.json with info from the package.json.  Leaves the dist, dist-modules and modules folders for supporting global, commonjs, and amd consumers.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,0 +1,35 @@
+{
+  "name": "react-dnd",
+  "version": "0.6.3",
+  "homepage": "https://github.com/gaearon/react-dnd",
+  "authors": [
+    "Dan Abramov <dan.abramov@me.com>"
+  ],
+  "description": "HTML5 drag-and-drop mixin for React with full DOM control.",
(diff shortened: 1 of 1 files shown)


gaearon:
How do I maintain this? Just increment a version number?
Why use Bower over NPM?

jgable:
Yes, just incrementing the version number should be fine as you publish.  You'll also need to [`bower register`](http://bower.io/docs/creating-packages/#register) it so it can be installed by name. 

> Why use Bower over NPM?

That's a good question, one that I'm sure the rest of the internet has a lot to say about better than I could.  In my current project we are setup to use Bower, and a lot of other components also support it.

If you'd rather not support it, I'm happy to fork and publish it myself.  If you ping me via an @jgable when a new version is published I'll also bump the bower fork.

gaearon:
I added you to collaborators, please feel free to bump it! I'll be pinging you on releases.

gaearon:
@jgable 0.6.4 is out

PR #43, opened 2015-02-03
fix DragDropMixin in firefox

nelix:
fixes https://github.com/gaearon/react-dnd/compare/experiments#diff-b16c9fe9f0a1d03ac2449ec2c5d339e3R266


nelix:
The coords mixin is maybe not with the style of the rest of the react-dnd code base, its adapted from code I had elsewhere. Do youhave an opinion on how what the correct way to do this?

gaearon:
I say let's get this in and tweak style later :-)

Anything else you feel lacking on that branch? I haven't played with it for so long I almost forgot what I wanted it to do lol

PR #44, opened 2015-02-03
use react-hot-loader and cli webpack-dev-server

nelix:
Use hot loader to make dev a little easier, also tried to simplify the config a little by using the cli webpack-dev-server


gaearon:
Need to split dev and production config, and turn off hot loader for `./scripts/build`.
Probably simplest way to go is using `process.env` variable and setting it in `./scripts/build`.

nelix:
I can have the whole project use a single webpack.config and perhaps a separate entry for examples? I just thought the  webpack.config.production was being used to bundle the examples so i didnt bother to check if the examples where in production mode

gaearon:
@nelix Let's have a separate config for examples if it simplifies things. I was just being lazy.

Issue #45, opened 2015-02-03
Clean up event handlers after unmounting 


gaearon:
As we move towards implementing alternative backends, we need to make `NativeDragDropSupport` clean up after itself when last draggable/droppable component is unmounted. Something like `setup()` and `teardown()`.

PR #46, opened 2015-02-03
clean up html5 event handlers after unmounting

nelix:
fixes: #45
Also renames the native support to /backend/HTML5 pending future backend support

### DIFF
--- a/modules/backends/HTML5.js
+++ b/modules/backends/HTML5.js
@@ -4,12 +4,12 @@ var DragDropActionCreators = require('../actions/DragDropActionCreators'),
     NativeDragItemTypes = require('../constants/NativeDragItemTypes'),
     DropEffects = require('../constants/DropEffects'),
     EnterLeaveMonitor = require('../utils/EnterLeaveMonitor'),
-    isFileDragDropEvent = require('./isFileDragDropEvent'),
+    isFileDragDropEvent = require('../utils/isFileDragDropEvent'),
     shallowEqual = require('react/lib/shallowEqual'),
     union = require('lodash-node/modern/arrays/union'),
     without = require('lodash-node/modern/arrays/without'),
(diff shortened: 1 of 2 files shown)


modules/backends/HTML5.js
```diff
@@ -49,58 +49,74 @@ function preventDefaultFileDropAction(e) {
...
+  }
 
-    triggerDragEndIfDragSourceWasRemovedFromDOM();
-  });
+  triggerDragEndIfDragSourceWasRemovedFromDOM();
 }
 
-var NativeDragDropSupport = {
+var HTML5 = {
+  begin() {
```
gaearon: Nit: I think `setup`/`teardown` describe what they do better, do you have any objections to this naming?

modules/backends/HTML5.js
```diff
@@ -49,58 +49,74 @@ function preventDefaultFileDropAction(e) {
...
+  teardown() {
+    if (typeof window !== 'undefined') {
+      window.removeEventListener('dragenter', handleDragEnter);
+      window.removeEventListener('dragover', handleDragOver);
+      window.removeEventListener('dragleave', handleDragLeave);
+      window.removeEventListener('drop', handleDrop);
+    }
+  },
+
   handleDragStart(dragTarget, imitateDragEnd) {
```
gaearon: It's slightly confusing that we have `handleDragStart` here (and it's called from outside), but `handleDragEnter` and friends are private and are in fact proper event handlers.

Can we find a more suitable name for `handleDragStart` and `handleDragEnd`? Something like `beginDrag` and `endDrag`, maybe.

gaearon: Another possibility is just using `handleGlobal*` naming for event handlers here.

gaearon:
:+1:

Issue #16, opened 2014-10-27
Touch backend


KyleAMathews:
It'd be killer if there was a touch backend that would be switched-in automatically for touch devices.

Georgette:
So I just had a request to make this drag and drop touch enabled. Has anyone done this and want to share tips?

gaearon:
@Georgette 

@nelix started work on `mousemove` backend, we'll do `touchmove` backend after that.

You can look at work in progress here: https://github.com/nelix/react-dnd/commit/8de7f7fe24c7ae397a971c517dada9323e6c27f0 (it's still far from being done and doesn't really work at this point). We still need to figure out which parts stay in `DragDropMixin` and which need to be moved into `backends/HTML5` (and other backends).

If you'd like to work on that, you can start with that commit and keep implementing mousemove backend. When it works, it shouldn't be hard to implement a similar touchmove backend.

gaearon:
Please, if you'd like to work on this, keep us updated here or in Gitter room, so we don't end up with two implementations.

gaearon:
Here's a proof of concept of touch support: https://github.com/gaearon/react-dnd/pull/143

You can play with it: http://rawgit.com/gaearon/react-dnd/touch-poc/examples/index.html

It uses HTML5 backend on desktop and touch backend on mobile.
Only [custom rendering](http://rawgit.com/gaearon/react-dnd/touch-poc/examples/index.html#/drag-around-custom) example will show a drag preview on mobile.

jareware:
This seems quite promising! Any ideas on when this might land? :)

gaearon:
I'm currently busy with another project and will be until July.
I might be able to work on this somewhere in July or August, but I'll be happy if somebody beats me to it :-)

flyon:
since 1.1 was mentioned in this thread and v1.1.4 is out. I'm just wondering if the proof of concept above is already included in 1.1.4?

gaearon:
Nope.

I shouldn't have used a precise version number cause you never know when you're going to bump :-)

I currently don't have time to work on this but the proof of concept is in the PR. I'll probably update the PR to reflect the current API and wait for somebody to actually implement it.

gaearon:
Let's continue the discussion in https://github.com/gaearon/react-dnd/pull/240.

gaearon:
FYI: https://github.com/yahoo/react-dnd-touch-backend.

haf:
[...]

PR #48, opened 2015-02-06
Implement mousemove backend

gaearon:
This is the first stab at an alternative backend. When we get this to a properly working state, implementing touch support (#16) will be a piece of cake.

- [x] Get item displaying while dragging in http://localhost:8080/drag-around-experimental
- [x] Get item dropping and changing position after drop in http://localhost:8080/drag-around-experimental 
- [x] Implement droppable (enter, leave, over)
- [x] Add barebone Touch backend
- [ ] Add a mousemove backend example for Dustbin and Sortable, add snapping to mousemove Drag Around
- [ ] Prevent text selection
- [ ] Make sure nested drop targets and drag sources work without a problem, just like with HTML5 backend
- [ ] Decide if we can infer (or allow injectable) mouse cursors for different `DropEffects`
- [ ] Figure out a way to set specific backends for components and have them exist on page without conflicts. Forbid operations on one backend if another backend is active. Components with different backends shouldn't notice each other. Think this through
- [ ] Make sure mouse is resilient against corner cases
- [ ] Make sure coordinates are calculated correctly regardless of scrolling position
- [ ] Implement Touch backend
- [ ] Internal refactoring (if needed)
- [ ] Update documentation

Huge thanks to @nelix for getting this ball rolling.


gaearon:
@nelix 

I made some progress, try `MouseDragDropMixin` and (quite broken) `TouchDragDropMixin`—both on container and item. `MouseDragDropMixin` should work OK now, although corner cases may be broken. `TouchDragDropMixin` only works for moving item and is horribly slow.

gaearon:
The useful part of this has already been extracted into 0.9 refactoring, and the rest will be superseded by future work on top of #111 so I'm closing this.

gavinwahl:
@gaearon Is there a mousemove backend now? I can't find documentation anywhere?

an5rag:
This seems promising: https://github.com/zyzo/react-dnd-mouse-backend

PR #50, opened 2015-02-07
fix initial drag position in experiment example

nelix:


modules/backends/Mouse.js
```diff
@@ -56,8 +56,8 @@ var Mouse = {
 
   getDragClientOffset(e) {
     return {
-      x: e.clientX,
-      y: e.clientY
+      x: e.clientX + window.scrollX,
```
gaearon: This kinda looks like: https://github.com/gaearon/react-dnd/blob/backend-mousemove/modules/utils/getMouseCoordinates.js#L3

That implementation should work better because not all browsers have `scrollX`/`scrollY` (I think FF doesn't)

PR #53, opened 2015-02-08
Nested sort example

nelix:
Has a bug where the drag target is not detected for empty columns (maybe a bug in the mouse backend)
Doesnt have a link yet
Column changing is not immutable


nelix:
see #65 65

Issue #54, opened 2015-02-08
Change API to make configureDragDrop static


gaearon:
I propose we make `configureDragDrop` static, and pass current `component` as the first argument to each method. This change is a stepping stone to fix `endDrag` not firing on unmounted component (#38, we'll pass `null` as `component`) and also a stepping stone to allow “resurfacing” drag sources (#53, we'll pass another component as `component`).

I'd like this change to be implemented in a backward-compatible way with a deprecation warning. We will make it a breaking change in 1.0 with #48.

- [ ] Support static `configureDragDrop` and pass `component` as the first argument to all methods
- [ ] Update the docs to use the new way
- [ ] Add a deprecation warning with upgrade instructions but still allow old-style `configureDragDrop` definitions (without `component` parameter)
- [ ] Make sure deprecation warning is logged once per class and includes class name (like React's own warnings)
- [ ] Prepare a separate PR that removes deprecation warning and makes it a breaking change. We will do that on `1.0` branch.

Any comments or volunteers?

@hakanderyal @nelix

gaearon:
To clarify. Before:

```js
var Card = React.createClass({
  mixins: [DragDropMixin],

  configureDragDrop(registerType) {
    registerType(ItemTypes.CARD, {
      dragSource: {
        beginDrag() {
          return {
            item: {
              id: this.props.id,
              children: <Card {...this.props}/>
            }
          };
        }
      },

      dropTarget: {
        over(item) {
          this.props.moveCard(item.id, this.props.id);
        }
      }
    });
  },
```

After:

```js
var Card = React.createClass({
  mixins: [DragDropMixin],

  statics: {
    configureDragDrop(registerType) {
      registerType(ItemTypes.CARD, {
        dragSource: {
          beginDrag(component) {
            return {
              item: {
                id: component.props.id,
                children: <Card {...component.props}/>
              }
            };
          }
        },

        dropTarget: {
          over(component, item) {
            component.props.moveCard(item.id, component.props.id);
          }
        }
      });
    }
  }
```

We may want to have a less nested API later, but this is not the point. The point is using `component` instead of `this` so we can later be free to pass `null` or another component if we wish so.

This breaks the use case when `this` is accessed directly in `configureDragDrop` but it's a rare use case and seems only useful in [contrived examples](https://github.com/gaearon/react-dnd/blob/master/examples/_dustbin-interesting/Item.js#L23).

gaearon:
Released in `react-dnd@0.6.4`.

PR #56, opened 2015-02-08
initial api changes implementing #54

nelix:


modules/mixins/DragDropMixin.js
```diff
@@ -54,6 +55,14 @@ function checkDropTargetDefined(component, type) {
   );
 }
 
+function callDragDropLifecycle(func, component) {
+  if (component._legacyConfigureDragDrop) {
+    return func.apply(component, rest(arguments, 2));
+  }
+
+  return func.apply(null ,rest(arguments, 1));
```
gaearon: Nit: wrong spacing

modules/mixins/DragDropMixin.js
```diff
@@ -157,8 +166,18 @@ var DragDropMixin = {
     this._dragSources = {};
     this._dropTargets = {};
 
-    invariant(this.configureDragDrop, 'Implement configureDragDrop(registerType) to use DragDropMixin');
-    this.configureDragDrop(this.registerDragDropItemTypeHandlers);
+    this._legacyConfigureDragDrop = !this.constructor.configureDragDrop;
+
+    if (this._legacyConfigureDragDrop) {
+      console.warn('configureDragDrop(registerType) should be defined as a static'); // TODO: make better?
```
gaearon: `Warning: <ClassName> declares configureDragDrop as an instance method, which is deprecated. Move configureDragDrop to statics and change signatures inside to accept component as first parameter.`

?

modules/mixins/DragDropMixin.js
```diff
@@ -203,7 +222,8 @@ var DragDropMixin = {
         this.constructor.displayName
       );
 
-      this._dropTargets[type] = defaults(bindAll(dropTarget, this), DefaultDropTarget);
+
```
gaearon: Nit: extra line?

gaearon:
This looks good. I'll give it a go on my main project and report back in an hour or so.

modules/mixins/DragDropMixin.js
```diff
@@ -157,8 +166,21 @@ var DragDropMixin = {
...
+        ' Move configureDragDrop to statics and change signatures inside to accept component as first parameter.',
+        this.constructor.displayName
+      );
+      invariant(this.configureDragDrop, 'Implement configureDragDrop(registerType) to use DragDropMixin');
+      this.configureDragDrop(this.registerDragDropItemTypeHandlers);
+    } else {
+      invariant(this.constructor.configureDragDrop, 'Implement configureDragDrop(registerType) to use DragDropMixin');
+      this.constructor.configureDragDrop(this.registerDragDropItemTypeHandlers);
+    }
+
```
gaearon: Nit: two extra lines

modules/mixins/DragDropMixin.js
```diff
@@ -157,8 +166,21 @@ var DragDropMixin = {
     this._dragSources = {};
     this._dropTargets = {};
 
-    invariant(this.configureDragDrop, 'Implement configureDragDrop(registerType) to use DragDropMixin');
-    this.configureDragDrop(this.registerDragDropItemTypeHandlers);
+    this._legacyConfigureDragDrop = !this.constructor.configureDragDrop;
+
+    if (this._legacyConfigureDragDrop) {
+      console.warn('Warning: "%s" declares configureDragDrop as an instance method, which is deprecated.' +
```
gaearon: I think it's important that we warn once per class instead of on every `componentWillMount`.
We can set `_legacyConfigureDragDrop` on the constructor and check if it's already been set before warning.

gaearon:
There's a small problem with [`DefaultDropTarget#getDropEffects`](https://github.com/nelix/react-dnd/blob/configureDragDrop-static/modules/mixins/DragDropMixin.js#L91). Since it's the same for both legacy and new-style drop target, the first parameter will either be `allowedEffects` or `component`.

It's just a single method so we can put a temporary hack there. For example, we can make it `LegacyDefaultDropTarget`, add a new one with modern signature and switch between them based on the legacy flag.

PR #57, opened 2015-02-10
Update bower package to 0.6.4

jgable:
I'll publish as soon as this is updated.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,6 @@
 {
   "name": "react-dnd",
-  "version": "0.6.3",
+  "version": "0.6.4",
   "homepage": "https://github.com/gaearon/react-dnd",
   "authors": [
     "Dan Abramov <dan.abramov@me.com>"


jgable:
@gaearon Next time, just need to bump the bower.json version before the release.  Otherwise, the bower.json in the .tgz will be different and give a little warning in the bower install:

screen shot 2015-02-10 at 9 30 48 am

Should still be good though.

gaearon:
Thank you! I'm sorry I forgot about it and only realized post factum :-(

gaearon:
@jgable 0.8 is out

jgable:
Everything is auto updating now on bower, as long as the bower.json version
is updated there won't be anything else you have to do.

gaearon:
Oh, that's sweet! How does it work?

jgable:
It matches the bower.json version with the github tags and downloads the .tgz files from Github directly; no need to publish them.  If you have bower installed you can check it out by running `bower info react-dnd`.

PR #59, opened 2015-02-10
dragSource tracking through unmount

nelix:
:dragon: :arrow_down:

### DIFF
--- a/modules/mixins/DragDropMixin.js
+++ b/modules/mixins/DragDropMixin.js
@@ -14,7 +14,6 @@ var DragDropActionCreators = require('../actions/DragDropActionCreators'),
     assign = require('react/lib/Object.assign'),
     defaults = require('lodash/object/defaults'),
     union = require('lodash/array/union'),
-    rest = require('lodash/array/rest'),
     without = require('lodash/array/without'),
     isArray = require('lodash/lang/isArray'),
     isObject = require('lodash/lang/isObject'),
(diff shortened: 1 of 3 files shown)


modules/mixins/DragDropMixin.js
```diff
@@ -71,7 +71,17 @@ function hashStringArray(arr) {
   return arr.join(UNLIKELY_CHAR);
 }
 
+var _nextDragSourceKey = 0;
+
 var DefaultDragSource = {
+  key() {
+    if (!this._dragSourceKey) {
+      // TODO: warn if key is missing
+      this._dragSourceKey = ('_' + _nextDragSourceKey++);
```
nelix: Maybe the displayName should be part of the key? Is there a way we could generate a key that works without consumer cooperation?

gaearon:
I think this fixes a different problem than #38. You're making sure `getDragState` makes sense even if drag source has “resurfaced” as a different component.

However, I think we need to (separately) fix #38 first. The problem there is following: `endDrag` won't get called on a component that was unmounted. Even though the backend will [“imitate”](https://github.com/nelix/react-dnd/blob/track-dragSource-though-unmount/modules/backends/HTML5.js#L42) drag end and call [`handleEndDrag`](https://github.com/nelix/react-dnd/blob/track-dragSource-though-unmount/modules/mixins/DragDropMixin.js#L322) on an unmounted drag source, [this check](https://github.com/nelix/react-dnd/blob/track-dragSource-though-unmount/modules/mixins/DragDropMixin.js#L330) won't let mixin call `endDrop` which, alas, seems desirable.

We might change

```js
    if (!this.isMounted()) {

      // Note: this method may be invoked even *after* component was unmounted
      // This happens if source node was removed from DOM while dragging.

      return;
    }

    this.setState({
      ownDraggedItemType: null
    });

    callDragDropLifecycle(endDrag, this, effect, e);
```

to

```js
    if (this.isMounted()) {
      this.setState({
        ownDraggedItemType: null
      });
    }

    callDragDropLifecycle(endDrag, this, effect, e);
```

This will get lifecycle method called. What I *don't* like about it is we'll be calling it on an unmounted component. This just doesn't feel right. But it's better than not calling it at all, I guess.

So I suggest we do this for now and then implement it more correctly with the breaking change.

Next steps will be:

* Remove support for deprecated API (jumping to 1.0)
* Change this logic to fire `endDrag` with `component == null` when source is unmounted
* Implement what you wrote in this PR on top of it, but so that `component` matches resurfaced component, and with warning if there is more than one mounted component with the same key

PR #64, opened 2015-02-11
Track drag source though unmount

nelix:


gaearon:
I think it's better to do this together with the resurfacing after #69.

nelix:
Mind if I leave the PR here to remind me?

gaearon:
Sure

gaearon:
TODO as I see it:

- [ ] Merge master into 1.0 (basically means reapplying changes from #67 after all the refactoring, SORRY for that happening again lol). Ideally 1.0 should be overwritten as a single commit over current master, that commit being similar to 91fd2c55671caf024addc0826a81afec080eb822.
- [ ] Rewrite this branch to be against 1.0 branch
- [ ] **Figure out** (see below) if we want to pass “resurfaced” component to `endDrag` or not
- [ ] Update Kanban example to use this branch

The upside of passing “resurfaced” component is that you can do something like `component.props.onEndDrag` without worrying that component may have resurfaced. For example, as Card is dropped into a different Kanban list, we might want to trigger some animation on that list.

The downside of doing that is sloppy behavior if there are two components with the same key on the screen at the same time. Which one would we pass? Does the most recently mounted one win? Why? Would we log a warning? Is there a legitimate use case for two components representing the same drag source? Should each of them receive `endDrag`? If one of them is “primary” and one of them “secondary”, which one would receive `endDrag`? Would both? What if there are zero components? (Source was removed and never added.) Should we pass `null`? Unmounted component? Finally, **what if resurfaced component has a different type?** Think `Card` that has a different type `WonderfulCard` in some other `List` in Kanban. We wouldn't want one component's `endDrag` to handle the other component, would we? But `WonderfulCard` might want to do something when the dragging ends. Should there be separate methods like `joinDrag`/`leaveDrag` for resurfaced drag sources?

gaearon:
So here's an alternative suggestion I just thought can be better than `getKey` stuff.

1. When dragging ends, we call `endDrag` with original component, no matter what
2. There is no `getKey`
3. *However* there are three new lifecycle methods: `canJoinDrag(component, item)`, `joinDrag(component, item)`, `leaveDrag(component, item, effect)`. (We should probably make `item` a second parameter in `endDrag` for consistency—a breaking change that should probably go with #93.)

Now let's define *a compatible drag source* as a component that registers a drag source with the same `type` as the one being dragged.

When something is dragged, here's what happens:

1. Component's `beginDrag` is called
2. For *each compatible drag source*, call `canJoinDrag(component, item)` where `component` is *that* component (not the one that is dragged!)
3. For those that returned `true`, call `joinDrag(component, item)` on them

Then, until the dragging ends:

1. Whenever another compatible drag source is mounted, `canJoinDrag` (and potentially `joinDrag`) are called on it as well
2. Whenever a compatible drag source that received `joinDrag` by returning `true` from `canJoinDrag` is unmounted, `leaveDrag(component, item)` is called on it
3. When dragging ends, original component receives `endDrag`
4. All compatible drag sources that received `joinDrag` by returning `true` from `canJoinDrag` will receive `leaveDrag(component, item, effect)` with the result drop effect

How does this API changes things? Instead of

```js
        dragSource: {
          getKey(component) {
            return DropTypes.COLUMN + component.props.id;
          },
```

you'd have

```js
        dragSource: {
          canJoinDrag(component, item) {
            return item.id === component.props.id;
          },
```

Let's do that?

gaearon:
There's a good stress case from #108: sortable example without keys. `isDragging` should work correctly even if components don't have keys and “receive” new items in `componentWillReceiveProps`.

gaearon:
`canJoinDrag`-like API is now implemented in dnd-core, so work there supersedes this PR.
https://github.com/gaearon/dnd-core/commit/62e7cd2ddc0d38207287a6c5ac4530aeb0374c5e

Issue #67, opened 2015-02-11
Remove support for deprecated API


gaearon:
Before proceeding with other issues like drag source key tracking, we need a PR against [1.0](https://github.com/gaearon/react-dnd/tree/1.0) branch that removes support for deprecated API.

@nelix Would you like to take this?

nelix:
reopening this one

gaearon:
Rewrote this on top of recent master in 1.0 branch. Thanks!

https://github.com/gaearon/react-dnd/commit/2147e8bb2e9ded39f7223c98ffbe1bdf5f30c6e4

PR #68, opened 2015-02-11
remove legacy api

nelix:
fixes: #67

### DIFF
--- a/modules/mixins/DragDropMixin.js
+++ b/modules/mixins/DragDropMixin.js
@@ -10,7 +10,6 @@ var DragDropActionCreators = require('../actions/DragDropActionCreators'),
     isFileDragDropEvent = require('../utils/isFileDragDropEvent'),
     bindAll = require('../utils/bindAll'),
     invariant = require('react/lib/invariant'),
-    warning = require('react/lib/warning'),
     assign = require('react/lib/Object.assign'),
     defaults = require('lodash/object/defaults'),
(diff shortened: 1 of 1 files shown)


modules/mixins/DragDropMixin.js
```diff
@@ -181,16 +158,7 @@ var DragDropMixin = {
     this._dragSources = {};
     this._dropTargets = {};
 
-    if (this.configureDragDrop) {
-      warning(
```
gaearon: `warning` seems unused now, we can remove it from `require` list on top

modules/mixins/DragDropMixin.js
```diff
@@ -181,16 +158,7 @@ var DragDropMixin = {
...
-        'Move configureDragDrop to statics and change all methods inside it to accept component as first parameter instead of using "this".',
-        this.constructor.displayName
-      );
-      this.constructor._legacyConfigureDragDrop = true;
-      this.configureDragDrop(this.registerDragDropItemTypeHandlers);
-    } else if (this.constructor.configureDragDrop) {
+    if (this.constructor.configureDragDrop) {
       this.constructor.configureDragDrop(this.registerDragDropItemTypeHandlers);
     } else {
       invariant(
```
gaearon: Don't really need this `if` and `else` branches. Just put `invariant` before `this.constructor.configureDragDrop` call.

PR #79, opened 2015-02-17
Allow customizable live drag preview

gaearon:
This will allow easy customizable live drag preview even with HTML5 backend.
We will expose `DragLayerMixin` and add `context` as second parameter to `configureDragDrop`.

Checklist: https://github.com/gaearon/react-dnd/issues/78
Milestone: 0.8


nelix:
I did a couple of things and updated the checklist. I might have some more time later tonight. Maybe.

PR #92, opened 2015-02-24
allow React >=0.13.0-beta.1

deepsweet:
Just like in [react-hot-loader](https://github.com/gaearon/react-hot-loader/blob/master/package.json#L11) to be able to use [`0.13` features](https://facebook.github.io/react/blog/2015/01/27/react-v0.13.0-beta-1.html) on project right now w/o peerDependencies hell. Looks like everything is working fine, API is backwards compatible.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "lodash": "^3.1.0"
   },
   "peerDependencies": {
-    "react": "0.12.x"
+    "react": "0.12.x || >=0.13.0-beta.1"
   },
   "devDependencies": {
     "6to5": "^3.5.3",


gaearon:
Good point. Can you verify that this includes recently released [v0.13.0-rc1](https://github.com/facebook/react/commit/06126ad3f4d063e89b3168abce79c9cd9961831c)? If not, can  you please make sure examples work with it, and if they do, add another `>=` separately?

deepsweet:
`>=0.13.0-beta.1` includes `0.13.0-rc1` as well. all examples are working fine for me, but not the react-router (https://github.com/rackt/react-router/issues/638) :)

PR #97, opened 2015-02-27
Support for native url drag and drop

sebcante:
hi @gaearon ,

That would be great to add support for native links to this library. This will allow user to drag links from other browser windows for example.

Enhanced from PR https://github.com/gaearon/react-dnd/pull/94
* fix formatting
* added IE support and make it more robust to null returns by dataTransfer.getData
* excluded `dist/` ,`dist-modules/`, `.gitignore` commits

Hoping we could merge this to master soon.

thanks! 

-seb


gaearon:
Thanks a lot, great job! Landed in 0.9.1.

PR #98, opened 2015-02-27
Add simpler nested example as a nested dropzone playground

gaearon:
This is based on https://github.com/gaearon/react-dnd/pull/90, but with some unneeded code removed and TODOs marked.


gaearon:
What I don't like about this approach: `isOverCurrent` is OK, but what about `enterCurrent`, `overCurrent`, `leaveCurrent`? These seem necessary, but I wouldn't like to make API twice as big.

On the other hand, for most use cases drop zones are probably either “deep” (care about themselves + their children) or “shallow” (care about just their area).

What if something like `shallow: true` was a flag on drop zone's definition, which would change the behavior of both `getDropState(type).isOver`, `enter`, `leave` and `over`? Could this solve proliferation of API methods?

@nelix @itrelease

nelix:
Making some changes to this based on #103 I think would clean up that render function and help make it a little clearer how it works.

gaearon:
Which render function are you referring to? The original commit was needlessly complicated, I simplified it and it should be straightforward now, is it not?

gaearon:
Superseded by dnd-core and `context.isOver(handle, shallow)` API.
We're figuring out how to transition to dnd-core in #111.

gaearon:
Need to port this to the new API and get this in.

gaearon:
Merged to `dnd-core`.

PR #85, opened 2015-02-17
Add nested example

itrelease:

PR #99, opened 2015-02-27
Capture dropzones for correct enter/leave state

gaearon:
This is one attempt at implementing `isOverCurrent` by @itrelease extracted from #85.

This still doesn't solve `enter` / `over` / `leave`. We want to be able to execute side effects on `enterCurrent`, `overCurrent` and `leaveCurrent`. (By the way I'm open to better name suggestions.)

If there are better API ideas, again, I'm all ears.

### DIFF
--- a/modules/actions/DragDropActionCreators.js
+++ b/modules/actions/DragDropActionCreators.js
@@ -33,6 +33,20 @@ var DragDropActionCreators = {
     DragDropDispatcher.handleAction({
       type: DragDropActionTypes.DRAG_END
     });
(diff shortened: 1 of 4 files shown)


gaearon:
For some potentially better API ideas: https://github.com/gaearon/react-dnd/pull/98#issuecomment-76321732

gaearon:
Superseded by dnd-core and `context.isOver(handle, shallow)` API.
We're figuring out how to transition to dnd-core in #111.

PR #102, opened 2015-02-27
allow data types to be represented as ES6 symbols

mistadikay:
With new “symbol” primitive type in ES6 there is a good pattern using symbols as constants values to guarantee these values to be unique. More about symbols here: http://www.2ality.com/2014/12/es6-symbols.html

Usage example:
```js
module.exports = {
  BLOCK: Symbol(),
  IMAGE: Symbol()
};
```


gaearon:
That's nice. Thanks!

gaearon:
Released in 0.9.1.

PR #111, opened 2015-03-08
Use DnD-core

nelix:
We're ready to go ahead with this.
dnd-core already has everything we need except for `over` hook which we'll add soon.

Things to be considered:

- [x] How do we change drag source / target API?
  - [x] We want to remove `enter` and `leave`
  - [x] Do we really want drag source / drop target API to differ from dnd-core? In some ways, it should differ because we want to be able to specify things like `dragPreview` that are specific to React DnD. Instead of passing `context` and `handle`, we want hooks to accept `context` and `component`, etc. So they're clearly not the same thing. How are they constructed? dnd-core prefers classes (or rather makes them easy to work with), React DnD currently uses plain objects. We need to decide that. We'll probably stick with wrapping sources/targets provided to React DnD into some dnd-core style wrappers before passing them to dnd-core. Similarly, `context` passed by React DnD to hooks will likely wrap dnd-core's context.
  - [x] Exotic things like `dragPreview`, `effectsAllowed`, etc. I need them to work (I actually use them). Need to figure out the right API for them.
 - [x] How many `collect*` methods do we need? Is it possible to pass them to mixin in its declaration instead of defining them on component? Can we do the same for registration?
- [x] Support native drag sources
- [x] How do we solve coordinate tracking? See https://github.com/gaearon/dnd-core/issues/6 for thoughts on this. We might come back to this when everything else is ready.

Overall the right course is to start from scratch and make examples work, from the simplest one to more complex ones.


gaearon:
I liked the idea of re-implementing examples from the ground up and then figuring out the new API.

Here's something I came up with just to get started:

```js
'use strict';

import React, { PropTypes } from 'react';
import { DragSource, DragDropManager } from 'dnd-core';
import ItemTypes from './ItemTypes';

class ItemDragSource extends DragSource {
  constructor(component) {
    this.component = component;
  }

  beginDrag() {
    return {
      name: this.component.props.name
    };
  }
}

class HTML5Backend {
  constructor(actions) {
    this.actions = actions;
  }

  getDraggableProps(sourceHandle) {
    return {
      draggable: true,
      onDragStart: (e) => this.handleDragStart(e, sourceHandle),
      onDragEnd: (e) => this.handleDragEnd(e)
    };
  }

  handleDragStart(e, sourceHandle) {
    this.actions.beginDrag(sourceHandle);
  }

  handleDragEnd() {
    this.actions.endDrag();
  }
}

const manager = new DragDropManager(HTML5Backend);

const style = {
  border: '1px dashed gray',
  backgroundColor: 'white',
  padding: '0.5rem',
  margin: '0.5rem',
  maxWidth: 80
};

const Item = React.createClass({
  propTypes: {
    name: PropTypes.string.isRequired
  },

  getInitialState() {
    return this.getDragState();
  },

  getDragState() {
    return {
      isDraggingItem: this.sourceHandle && manager.getContext().isDragging(this.sourceHandle)
    };
  },

  componentWillMount() {
    this.sourceHandle = manager.getRegistry().addSource(ItemTypes.ITEM, new ItemDragSource(this));
  },

  componentDidMount() {
    manager.getContext().addChangeListener(this.handleDragContextChange);
  },

  componentWillUnmount() {
    manager.getRegistry().removeSource(this.sourceHandle);
    manager.getContext().removeChangeListener(this.handleDragContextChange);
  },

  handleDragContextChange() {
    this.setState(this.getDragState());
  },

  render() {
    const { name } = this.props;
    const { isDraggingItem } = this.state;
    const opacity = isDraggingItem ? 0.4 : 1;

    return (
      <div style={{ ...style, opacity }}
           {...manager.getBackend().getDraggableProps(this.sourceHandle)}>
        {name}
      </div>
    );
  }
});

export default Item;
```

It's dirty but at least gets the `Item` highlight and unhighlight. I guess we can move on with something like this, cherry-picking parts of current backend code until it takes shape.

nelix:
I finished off the first example, I was not sure if commiting it was a good idea... But I figure you can just commit over the top of it :P

gaearon:
To clarify this work: we want to reimplement the examples with vanilla core-dnd, and *then* figure out which parts should be extracted into React DnD (most probably HTML5 quirk handling, some state helper mixin or component). We will definitely offer both a mixin and some easy way to use this from ES6 classes.

gaearon:
I wrote an article on the future directions for React DnD: [The Future of Drag and Drop APIs](https://medium.com/@dan_abramov/the-future-of-drag-and-drop-apis-249dfea7a15f).

raysuelzer:
When this gets pulled is it safe to assume that it will cause breaking changes?

gaearon:
@raysuelzer 

Yes, there will be breaking changes. It's a major overhaul. The migration path should be straightforward though because many restrictions are removed, not added.

This will depend on React 0.14 so it's only coming after 0.14 is out.

ockham:
> This will depend on React 0.14 so it's only coming after 0.14 is out.

I'm interested in trying out react-dnd. With the imminent migration to dnd-core, I'd rather avoid undergoing a predictable API change this big in the near future. OTOH, the product I'd need react-dnd for is currently still based on React 0.12 and ES5. We're probably going to migrate to React 0.13 and ES6 within the next couple of weeks (well, maybe months), but not right now.

Given these circumstances, is there any procedure you'd suggest for adopting react-dnd? (On a sidenote, why is react-dnd going to require React 0.14 anyway? Can we do without if we don't need a particular feature?)

gaearon:
@ockham I changed my mind, it will only require 0.13. You can already try 1.0.0-alpha.2 and look into examples. It doesn't have docs yet; I'm on it, expect them to land in a week. Examples are ES6/7-heavy but it's totally usable from ES5.

gaearon:
You should be able to migrate to React 0.13 rather painlessly. It doesn't change much, just introduces some migration/deprecation warnings.

ockham:
Awesome, I'll try it out!

PR #114, opened 2015-03-10
nulls offsets on drag end and exposes initial container offset to context

warp:
I am using the initial offset from container to calculate offset when dragging between containers.


gaearon:
:+1:

gaearon:
Released in 0.9.5

PR #126, opened 2015-03-24
Item returned in beginDrag should be an object

sgwilym:
Just a little thing! Tripped up on this myself while familiarising myself with react-dnd, and kept hitting
`'Error: Invariant Violation: Expected return value of beginDrag to contain "item" object.` 

It's obvious rereading it now, but I think this might save the kinds of people who need this walkthrough (e.g. _myself_) from being discouraged from taking advantage of your hard work. :smile:


gaearon:
That's good. Thank you.

PR #129, opened 2015-03-27
Wrap checkboxes and text with <label>

olegafx:
Examples (“drag-around-naive” and “drag-around-custom”)

Before: only checkbox itself is clickable
Now: checkbox and it’s label are clickable


gaearon:
Woah, thanks!

PR #131, opened 2015-04-01
Add Mouse backend that tracks mouse enter/leave based on mouse point pos...

agrauch:
...ition

Mouse backend is based on the branch backend-mousemove. 

After testing the MouseDragDropMixin in various browsers, we notice several problems:
* MouseEnter/Leave events were unreliable in Chrome 41
* MouseUp events were swallowed by a drag layer unless 'pointer-events: none' was added to the drag layer styles. Unfortunate this property is not supported by IE9/10

This implementation does not add mouse events to the drop target. Instead the MouseMove handler looks for drop targets based on pointer position. Drag enter/leave are triggered when a drop target is found and the drop target changes. 

This has some performance implications, as more drop targets are added the time to find a target will increase. It also relies on some React internal properties to determine if drop targets should be tracked and could potentially break with future React releases.


gaearon:
Wow, that's cool! Thanks. I'm not merging this now because we're in the process of a [big rewrite](https://github.com/gaearon/react-dnd/pull/111) for the next version. I'll come back to this after it's done (might take several more weeks).

agrauch:
Cool. Hope the rewrite goes well and this can get after it's done. We're looking forward to the new version.

Thanks.

gaearon:
If you'd like, you can try porting this on top of `dnd-core` branch. It's a big rewrite, so backend interface is somewhat different, but it's much more flexible and completely decoupled from React.

Basically, new backends have this interface:

```js
class MyBackend {
  constructor(manager) {
    // you will use actions to tell the engine that
    // drag started, hover, drop or something else occured:
    this.actions = manager.getActions();

    // you can use monitor to check current state,
    // like what item and type is dragged, whether we can drop or not:
    this.monitor = manager.getMonitor();

    // you can use registry to add drag source and drop targets
    // if your backend has any “built-in” sources or targets:
    this.registry = manager.getRegistry();
  }

  setup() {
    // will be called whenever there are > 0 components using this backend
    // you might want to set up some top-level handlers here
  }  

  teardown() {
    // will be called whenever there are no more components using this backend
    // you might want to remove some top-level handlers here
  }  

  connect() {
    // what "roles" does your backend provide?
    // these will be available as connect.* when configuring the component

    return {
      // these two make sense for any backend:
      dragSource: this.connectSourceNode,
      dropTarget: this.connectTargetNode,

      // this one is optional. you can also invent your own if you're into exotic stuff :-)
      // dragSourcePreview: ...
      // myWeirdThing: ...
    };
  }

  connectSourceNode(sourceId, node, options) {
    // this method "connects" a drag source node to this backend.
    // here's an EXAMPLE of what you could do:

    // you might want to set some attributes and bind some events
    const handleDragStart = (e) => this.handleDragStart(e, sourceId);
    node.setAttribute('draggable', true);
    node.addEventListener('dragstart', handleDragStart);

    return () => {
      // return a function that undoes everything you just did!
      node.removeEventListener('dragstart', handleDragStart);
      node.setAttribute('draggable', false);
    };
  }

  connectTargetNode(targetId, node, options) {
    // again, it's up to you. do WHATEVER set up for this node.
    
    // just be a good person and return a function that undoes this
    return () => {
      // clean up
    };
  }

  connectMyWeirdThing() {
    ..
  }
}

// you're supposed to export a function
export function createMyBackend(manager) {
  return new MyBackend(manager);
}
```

Check out the [new HTML5 backend](https://github.com/gaearon/react-dnd/blob/dnd-core/modules/backends/HTML5.js#L61) for an extensive example.

Here's a pattern it uses a lot:

* Attach top-level capture and bubble handlers;
* Attach individual handlers to nodes;
* As event bubbles, memorize source/target IDs whose nodes received the event;
* In the top-level bubble handler, emit an action using [`this.actions.something()`](https://github.com/gaearon/react-dnd/blob/dnd-core/modules/backends/HTML5.js#L412).

To use an arbitrary backend, just pass it to `configureDragDropContext` on a top-level component [like here](https://github.com/gaearon/react-dnd/blob/dnd-core/examples/_dustbin-simple/Container.js#L25).

agrauch:
Thanks for the documentation of the new interface. I'll take a look at getting this rewritten.

gaearon:
I'm going to leave 1.0 mostly as is for the next month so I have time to prepare my conf talk.
Adding touch (and possible mouse) backends will be a priority for 1.1 that should land somewhere in July.

Thanks for your work!

limscoder:
@gaearon We're would like to get up-to-date with react-dnd and get off of our custom fork. We would like to take a look at porting our mouse backend over to the current version of react-dnd.

One thing I noticed is that we are using DragLayerMixin from react-dnd, but we are moving to ES6 classes that don't support mixins. Is there an alternative in the current version of react-dnd?

What can we do to help ensure whatever code we come up with is mergable into react-dnd?

thanks!

gaearon:
>One thing I noticed is that we are using DragLayerMixin from react-dnd, but we are moving to ES6 classes that don't support mixins. Is there an alternative in the current version of react-dnd?

There is a higher-order component for that: http://gaearon.github.io/react-dnd/docs-drag-layer.html

limscoder:
@gaearon if we work on a Mouse backend that works with the latest version of react-dnd, will you help us get it merged in?

gaearon:
@limscoder 

I think it's better off as a separate package on NPM (we'll link to it from the docs). I'm probably going to move HTML5 backend to separate package too, so it can be versioned separately and forked more easily.

gaearon:
(See also [touch backend](https://github.com/gaearon/react-dnd/pull/240) as an inspiration)

limscoder:
I feel like a separate repo is a lot harder to maintain, because breaking pushes to master won't fail tests related to a backend living in a separate repo.

gaearon:
>I feel like a separate repo is a lot harder to maintain, because breaking pushes to master won't fail tests related to a backend living in a separate repo.

We use semver. If there's a breaking change, we'll bump major version. If there's a bug, we'll fix it.
Separate repo is the way to go here.

>It would be very easy to end up in the same situation we are in now

It wouldn't. The backend would enforce compatible `react-dnd` version via a peer dependency.

I'm closing because I'm confident I want backends to be in separate repos.
Please let us know in this thread if you publish one.

gaearon:
I propose that we use `react-dnd-*-backend` naming scheme, e.g. `react-dnd-mouse-backend`.

limscoder:
Semver doesn't really solve our problem. Our problem is that we have code that works with an old version of react-dnd, and it's going to be expensive to port to a new version of react-dnd.

We're shy of spending money to update our backend and then having another major version change come down the pipe in a month or two that totally breaks us and puts us back into the spot of being stuck on an out-of-date version of react-dnd.

Having the backend in the same repo would fix this problem, because maintainers would have to take the backend into account as they make changes. A major release would be forced to update any backends, because otherwise tests would fail.

I understand that this adds significant maintenance overhead to you and other core maintainers for functionality that you may not care about, so we will investigate our options. Our current code works fine, but we are stuck on React.createClass for any components that need DnD. I will let you know if we decide to create a react-dnd-mouse-backend.

Also, on a not really related note, thanks for your work on Redux. Hawt. Pure reducers are so simple and easy to test, especially when using immutablejs. I really like it.

gaearon:
>We're shy of spending money to update our backend and then having another major version change come down the pipe in a month or two

Please understand that there are no such plans.
1.0 was a major rewrite, and it took me several months.
I'm not planning to do any other major rewrites of React DnD.

limscoder:
Ok, good to know.

gaearon:
FYI: #282, let me know if you are interested.

methyl:
@gaearon @limscoder @agrauch react-dnd-touch-backend just got mouse support: https://github.com/yahoo/react-dnd-touch-backend/pull/9

limscoder:
Sweet, procrastination pays off again...

PR #136, opened 2015-04-10
Update bower.json to allow React 0.13.x.

trevorsmith:
Using the same dependency range specified in package.json.


gaearon:
Thanks. Do I need to bump the version to get this in?

trevorsmith:
I don't think you'll need to, React 0.13.1 is now being accepted as a dependency on my end.

gaearon:
Great!

PR #153, opened 2015-05-15
Document `getInitialOffsetFromContainer()` in API doc

MadLittleMods:
Add `DragDropContext.getInitialOffsetFromContainer()` to the API documentation.

Currently it is a "hidden" method that needs documentation.

I just had a use for it today and stumbled upon it while logging the `DragDropContext`

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -119,6 +119,10 @@ Returns `{ x, y }` coordinates of the last recorded mouse position in client coo
 
 Returns relative `{ x, y }` coordinate delta between last recorded mouse position and its position when dragging started.
 
+##### `getInitialOffsetFromContainer()`
+
+Returns `{x, y}` coordinates of mouse position relative to the top-left corner of the drag source element.
+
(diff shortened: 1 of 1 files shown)


gaearon:
It's actually removed in 1.0, can you give me more info about your use case? I could probably bring it back in 1.1.

MadLittleMods:
@gaearon My use case is to position the element a newly created element right where it was dropped. didn't want to deal with the delta because looking up the original drag source is not as straight forward.

Here is a demo with it in action: http://codepen.io/MadLittleMods/pen/VLaENY

is:
@MadLittleMods you can try html5 dom getBoundingClientRect, it's dom node position based on the view offset. it's comparable with DropMonitor.getClientOffset.

MadLittleMods:
@is You can use the demo linked in a previous message to try it out.

gaearon:
If you're still interested in this feature, please create a PR adding it.

PR #159, opened 2015-05-20
Proof of concept text data transfer support

### LINKED ISSUE #158: IE bookmarklet link drag disabled
So this is a pretty specific issue. It affects only IE (tested in IE10 & IE11) and only [bookmarklet](http://en.wikipedia.org/wiki/Bookmarklet) links. 
While using React DnD all bookmarklet links become completely undraggable. Users are unable to drag them to the Favorites bar. Normal links are unaffected. 

Problem might be related with the 'dragstart' listeners in the HTML5 backend. (Tested by commenting them out, which fixed that problem but ofc broke the plugin)

CodePen to showcase the problem: http://codepen.io/npredovan/pen/Kpgdvo

gaearon:
This is intended to fix #158 and also add support for registering drop targets for text (e.g. from input boxes).

### DIFF
--- a/examples/01 Dustbin/Multiple Targets/Container.js
+++ b/examples/01 Dustbin/Multiple Targets/Container.js
@@ -35,6 +35,8 @@ export default class Container extends Component {
 
     return (
       <div>
+        <a href='javascript:alert("hi")'>Drag me!</a>
+
         <div style={{ overflow: 'hidden', clear: 'both' }}>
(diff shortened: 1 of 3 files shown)


gaearon:
Closed in favor of 6d54bb55b0e3eaabbc88c05554bac9cbcb3777b6, which is released as of 1.1.

PR #190, opened 2015-06-12
Add minified builds back to NPM/Bower

paulruescher:
In its current state, the Bower package is a docs folder and some config files. Adding the minified version back covers corner cases (e.g. AMD).

Discussion found in https://github.com/gaearon/react-dnd/issues/189#issuecomment-111467063

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,5 +1,4 @@
 src
-dist
 scripts
 examples
 site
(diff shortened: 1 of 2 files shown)


gaearon:
:+1: 

Should I bump the patch versions now?

paulruescher:
That'd be great :ok_hand:

PR #201, opened 2015-06-19
Added a note about `clientOffset` to the testing tutorial

JustinTulloss:
It was a little bit difficult to figure out how to pass around client
offsets, so this patch adds a guiding comment to the example test.


gaearon:
:+1:

PR #214, opened 2015-06-27
Dustbin simple ES5 example

vkbansal:
Simple Dustbin example inES5 for issue #200

PR #216, opened 2015-06-28
ES5 port of ReactDnD Sortable Simple example

awarberg:
The Sortable Simple example was very useful to me so I wanted to contribute something by porting it to ES5. I don't know how to work the build system and I wanted to provided a runnable example so this is the best I could come up with.

### DIFF
--- a/es5-sortable-simple-demo.html
+++ b/es5-sortable-simple-demo.html
@@ -1,0 +1,203 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <title>React DnD ES5 example for Sortable Simple</title>
+    <meta http-equiv="Content-Type" content="text/html; charset=utf-8"/>
+    <meta http-equiv="X-UA-Compatible" content="IE=edge"/>
+    <meta name="description" content=""/>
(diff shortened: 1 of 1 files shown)


gaearon:
Thanks a lot! The way I see it, maybe we should set up a separate repo like `react-dnd-es5-examples`? This way they can have a different (much simpler) build pipeline (or no build at all), and we avoid some confusion.

This repo will be linked to from the `examples` directory of React DnD so it stays visible. We'll also add a special “ES5 examples” page to the docs that links to this repo.

I'm open to anyone maintain that repo—for example, @vkbansal who already did some work with #214 too. Does this make sense?

vkbansal:
I think that would be a good idea. It can be something simpler and similar to what you had for pre-1.0.0

gaearon:
If anyone's up for creating it please do and transfer these PRs there. I'll edit the docs then.
Thanks!

gaearon:
@awarberg Can you please move this to https://github.com/vkbansal/react-dnd-es5-examples? Thanks!

PR #225, opened 2015-07-09
Swapping parameters in overview docs code snippet

Anifacted:
The parameters seem to be reversed here, no?


gaearon:
Indeed, thanks!

PR #251, opened 2015-07-28
Simple sortable example: hover() can handle different heights of elements

kraftwer1:


gaearon:
Awesome! I'll test it as soon as I get some time after releasing Redux 1.0. That's where my time was going the past few weeks.

examples/04 Sortable/Simple/Card.js
```diff
@@ -17,12 +17,42 @@ const cardSource = {
...
     const draggedId = monitor.getItem().id;
-
-    if (draggedId !== props.id) {
-      props.moveCard(draggedId, props.id);
+    if (draggedId === ownId) {
+      return;
+    }
+  
+    // assuming something like findTodoIndex is implemented
+    const ownIndex = props.findTodoIndex(ownId);
```
gaearon: Pretty sure there's something missing: (1) there is no such method in props, (2) this comment doesn't need to be in source, it was only for that Github discussion. :-)

kraftwer1:
You're totally right! I've updated it. But in my opinion this whole thing is too low-level to have the users implement it on their own. I'd rather see it in the core of React-DnD. What about you?

gaearon:
User can always encapsulate this in component.
All React DnD provides are bindings for drag and drop interaction—not specific set of controls.
I think being low level is fine in this case.

kraftwer1:
Then let's hope the example helps people. Is it running now, can you merge it?

gaearon:
I will, after dealing with #277.

gaearon:
This is in, great job!

PR #254, opened 2015-07-29
Also pass this.component to canDrop callback

delijah:


gaearon:
We don't do this for a reason. It won't be available in some cases where a component is just being mounted. Why aren't `props` enough in your case?

delijah:
you are right, i found the doc entry for this.

well the problem is connected to the following issue: https://github.com/gaearon/react-dnd/issues/233

i want to check if the user is allowed to drop the current dragged element within the canDrop function. to check it, according to your example, i need the boundingClientRect of the component.

gaearon:
Rather than introduce an inconsistency, I'd encourage you to have two separate drop targets whose `canDrop` is provided as a prop by a parent undroppable drop target. It may set the state from `hover` method, for example.

PR #255, opened 2015-07-29
Typo fix.

volrath:
Example is showing how to make a ``Cell`` a ``DropTarget``, but description talks about "drag source".

Maybe the change has to be the other way around? maybe we need to change the example code? Although the change will be bigger, we'd need to change ``isOver`` and ``canDrop`` for some ``DragSourceMonitor`` methods. Let me know.


gaearon:
:+1:

Issue #221, opened 2015-07-02
Firefox is navigating to dragged image URL on drop


Enet4:
For some reason, FF's default behavior was not prevented in my project. I have a drag source `ImageItem` containing an image, and dropping it over my drop target will not only trigger my logic in the web application, but also issue a navigation to the image's URL.

> In Firefox 3.5+, the default behavior for a drop event is to navigate to the URL that was dropped on the drop target. That means dropping an image onto the drop target will result in the page navigating to the image file ([ref](http://www.developerfusion.com/article/144828/the-html5-drag-and-drop-api/))

I may provide more information on request. The examples appear to work fine. I wish I had an SSCCE of the problem... This could also be related to #219.

gaearon:
This indeed seems like a duplicate of #219.
I don't currently have time to work on this but I'm happy to provide guidance if you'd like to fix it.

gaearon:
Fixed in https://github.com/gaearon/react-dnd/commit/2299ecea6f292a5814c7efd6ae3460066b6b8707.

gaearon:
The fix for this issue was released in `react-dnd@1.1.8`.
Please verify that it works.

thomasvm:
I had this very issue, and I can confirm that this is fixed after upgrading to react-dnd@1.1.8

PR #259, opened 2015-08-06
fix selected text drop issue in Firefox #219

ttrmw:
Moved `e.preventDefault()` outside of the `isDraggingNativeItem` check, as the default must be prevented in Firefox for non Native items too.

Should `isFirefox` be checked here? I tested in Chrome and found no issue with preventing the default.


ttrmw:
@gaearon Any movement on this? Looks like it's not quite a dupe of #242 as the default must be prevented at capture AND bubble for our separate issues

gaearon:
Hey, sorry, I don't currently have time to review this. I might have some time in September.

As you can see from React DnD API, the backend is pluggable.
Feel free to fork HTML5 backend for now, publish it as a separate package, and use it with your fix.

ryanhiebert:
I can confirm that this works for me. I was seeing the issue mentioned in #221, and using this patched solved the problem.

ttrmw:
Hey @ryanhiebert, I think #221 is subtly different. @Enet4 found my patch didn't help. 

There is another PR -  #242 - which cancels Firefox's event handling at bubble. That fixed #221 for Enet4 but didn't help me with #219

ryanhiebert:
It may be subtly different, but it did solve my case, which is what was described in #221. I found that if I dropped an avatar which was a part of my drag layer, it would open the url to the image. Using your patch, I no longer have that issue.

gaearon:
Thank you very much for the PR. Your PR did not merge cleanly because you included the dist file—there is no need to do this in PRs. I cherry-picked your change as a different commit so I'm closing this via 2299ece.

Cheers!

gaearon:
Please verify that it works.

ttrmw:
Hey @gaearon - yep, works as expected. Sorry about the updated dist & thanks for the sort-of-merge!

PR #272, opened 2015-08-31
fix selected text drop issue in Firefox #219

### LINKED ISSUE #219: Firefox attempting erroneous requests on drop if dragSource text highlighted (all OSes)

### LINKED ISSUE #221: Firefox is navigating to dragged image URL on drop

ryanhiebert:
This commit by ttrmw should fix #219 as well as fix #221. I've tested this myself with the issue that was described in #221, which has a note pointing out that it seems to be a duplicate of #219. This is the same commit as previously mentioned on #219, except rebased to v1.1.5.


ryanhiebert:
Sorry, totally missed #259.

gaearon:
Please verify that it works.

PR #288, opened 2015-09-23
Improved docs (consistent naming, improved performance)

kraftwer1:


gaearon:
Thanks.

kraftwer1:
You're welcome. I was wondering about your opinion on that:

```javascript
// Update the monitors item with the new index
monitor.getItem().index = hoverIndex;
```

It's destructive, maybe there's a cleaner way. But I had the issue a couple of times when I had to make updates on the element during its dragging lifecycle.

gaearon:
I'll leave a special comment there.
It's mutative but it's exactly the case where you want to trade mutation for performance.

PR #318, opened 2015-10-29
A generic sortable example and components for Issue 305

globexdesigns:
Hey @gaearon,

Thanks a lot for your work on this project and for your time in helping us figure out a proper upgrade solution from 1.0 to 2.0.

This PR is an example which demonstrates our usage of react-dnd and the challenge reported here: https://github.com/gaearon/react-dnd/issues/305#issuecomment-152247354

Steps to reproduce:
- Run `npm run start`
- Navigate to the new "Generic" example page under "Sortable"
- Here we are showing a simple example of our `<Sortable/>` component whose job is to accept any element and make it sortable. The example enabled by default is using <div> elements and is working just fine. However, if you go into the code, you'll see we also try to sort `<CustomElement/>` object -- and this isn't working.

If you take a look at our `<SortableItem/>` component you'll see some comments in the render methods as to what we tried and what isn't working.

Any recommendations would be greatly appreciated.

### DIFF
--- a/examples/04 Sortable/Generic/Sortable.js
+++ b/examples/04 Sortable/Generic/Sortable.js
@@ -1,0 +1,208 @@
+import React, {Component, PropTypes} from 'react';
+import {DropTarget} from 'react-dnd';
+import SortableItem from './SortableItem';
+
+const _getLastChildKeyFromComponent = (component) => {
+  const allItems = component.state.children;
+  let lastChild = null;
+  React.Children.forEach(allItems, (child, i) => {
+    if (i === React.Children.count(allItems) - 1) lastChild = child;
+  });
+  return lastChild.props.eventKey;
+};
+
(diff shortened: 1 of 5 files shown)


globexdesigns:
Closing this. The solution is to change:

```
return React.cloneElement(child, {
      className: classes.join(' '),
      ref: (instance) => connectDragSource(connectDropTarget(findDOMNode(instance)))
    });
```

To:

```
return React.cloneElement(child, {
      className: classes.join(' '),
      ref: (instance) => {
        const instanceDOM = findDOMNode(instance);
        connectDragSource(instanceDOM);
        connectDropTarget(instanceDOM);
      }
    });
```

I will open a separate PR to update the docs accordingly.

gaearon:
Yup, this will work. Glad you figured it out!

PR #319, opened 2015-11-01
Added component as parameter to canDrop callback

daniel-lundin:
Thank you for a very well-architectured library!

I've found that having the component as a third parameter to the canDrop-callback of DropTarget would be very useful. In our use cases we often want to ask the target component if drop should be possible.

Thanks!

### DIFF
--- a/docs/01 Top Level API/DropTarget.md
+++ b/docs/01 Top Level API/DropTarget.md
@@ -62,7 +62,7 @@ The second `spec` parameter must be a plain object implementing the drop target 
 
 * **`hover(props, monitor, component)`**: Optional. Called when an item is hovered over the component. You can check `monitor.isOver({ shallow: true })` to test whether the hover happens over *just* the current target, or over a nested one. Unlike `drop()`, this method will be called even if `canDrop()` is defined and returns `false`. You can check `monitor.canDrop()` to test whether this is the case.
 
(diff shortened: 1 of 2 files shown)


gaearon:
Can you please tell me more about your use case? Why is `monitor.getItem()` not enough to make this decision?

daniel-lundin:
monitor.getItem() would return the object returned by beginDrag , right? I want a reference to the component that DropTarget wraps.

My use case is that I only want to allow dropping one object on the component wrapped DropTarget. If an item has been previously dropped, canDrop should return false. Only the DropTarget's wrapped component knows if an item has been dropped there before so I would use it something like this:
```
canDrop(props, monitor, component) {
  return !component.havePreviouslyDroppedItem();
}
```

Am I making sense? :)

gaearon:
So you're saying it depends on the component state? You could do this instead:

```js
render() {
  let element = <div>stuff</div>
  if (!this.state.didDropBefore) {
    return this.props.connectDropTarget(element);
  }
  return element;
}
```

In general `component` reference is a necessary evil for `drop()` and `hover()` so you can cause side effects related to component, its DOM node, etc. However I'd rather avoid passing it anywhere else.

daniel-lundin:
Ah, that's much nicer. That would solve my problems.
Thank you!

gaearon:
No problem, glad to help!

chmanie:
I am trying to implement [your scrolling solution](https://github.com/gaearon/react-dnd/issues/324#issuecomment-154042432) on an element that's not the window. I have to stop the scrolling interval (cancelAnimationFrame) when the element was dropped.  The only way I can think of to get ahold of the underlying DOM element (to actually scroll it via setting `scrollLeft`, `scrollTop`) or the scroll frame interval would be via the component itself (which is not passed into `canDrop`). Do you have another idea on how to accomplish this?

PR #327, opened 2015-11-06
Fix DragLayer ES5 documentation example.

CarlRosell:
Fix small mistake where ES5 example contained `const`.


gaearon:
Whoopsie, thanks.

PR #339, opened 2015-11-22
build both dev and minified versions to dist

johnswanson:
I'm building a Cljsjs package (https://github.com/cljsjs/packages/wiki/Creating-Packages) for react-dnd and would like a development version that I can minify during the build process (so that consumers of the package have access to both versions).

I don't have much experience with webpack, so I'm not sure if there's a good reason to use UglifyJs instead of Webpack's `--optimize-minimize` command-line option--if there is, I can make a new pull request.

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -2,7 +2,8 @@
 
 # build minified standalone version in dist
 rm -rf dist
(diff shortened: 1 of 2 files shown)


gaearon:
Thanks!

PR #348, opened 2015-12-19
Subscribe DragLayers to state changes. Fixes #293

### LINKED ISSUE #293: DragLayerMonitor().isDragging() doesn't get reset to false when stop dragging
isDragging is false when initialized. When dragging starts, it gets set to true. When the dragging stops, its value will still remain true.

ksonnad:
Fixes the issue noted in #293 by subscribing DragLayers to state changes from Monitor in addition to offset changes.  Now, ending a drag with a drop correctly calls the DragLayer's collect function.

### DIFF
--- a/src/DragLayer.js
+++ b/src/DragLayer.js
@@ -68,11 +68,13 @@ export default function DragLayer(collect, options = {}) {
 
       componentDidMount() {
         const monitor = this.manager.getMonitor();
-        this.unsubscribe = monitor.subscribeToOffsetChange(this.handleChange);
+        this.unsubscribeFromOffsetChange = monitor.subscribeToOffsetChange(this.handleChange);
+        this.unsubscribeFromStateChange = monitor.subscribeToStateChange(this.handleChange);
       }
 
       componentWillUnmount() {
-        this.unsubscribe();
+        this.unsubscribeFromOffsetChange();
+        this.unsubscribeFromStateChange();
       }
 
       handleChange() {
--- a/src/DragLayer.js
+++ b/src/DragLayer.js
@@ -68,11 +68,13 @@ export default function DragLayer(collect, options = {}) {
 
       componentDidMount() {
         const monitor = this.manager.getMonitor();
-        this.unsubscribe = monitor.subscribeToStateChange(this.handleChange);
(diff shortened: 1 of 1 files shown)


gaearon:
It seems like in most cases this will cause the handler to fire twice because there are two subscriptions.

I guess we can just leave `subscribeToStateChange()`.
Do you see any performance slowdowns from this?

ksonnad:
As far as I can tell and test, DragLayer's handler will only be called for changes it needs to know about, due to state checking in the Monitor's subscribe handlers.  It's true there are two of those now, but they're checking for different things that DragLayer needs.  I haven't noticed any slowdown, but my use case isn't particularly demanding.

ksonnad:
So when the state and offset change in the same cycle, DragLayer's handler will be called twice.  But I'm assuming state changes are relatively rare.

gaearon:
Aren't state changes superset of offset changes? I don't quite remember.

ksonnad:
I don't fully understand the code yet, I think you're right.  The offset values change even when subscribing only to the state change.  I'll amend the PR.

ksonnad:
Are there any other concerns?

gaearon:
No concerns. Sorry for delayed response. Please use a fork for now—I'll release an update as soon as I get some free time.

nickclaw:
Hey @ksonnad, your last two commit actually break custom drag layers. Here's what the custom drag layer demo looks like...
 
untitled

It looks like for some reason it only updates the offset when the mouse leaves or enters a another element. Everything seems to work fine if only the changes of your first commit are included.

ksonnad:
Glad you caught that, @nickclaw.  This is tricky for someone as new to the code as I am; it requires comprehension of dnd-core and html5-backend in addition to react-dnd.

I've reverted the offending commits (and can create a new PR if a cleaner history is desired).  Performance profiling of the changes reveals no noticeable regressions.

arcanis:
@gaearon I'd be really interested for this PR to be merged and published

@ksonnad Not really important (especially if it delays this PR) but should you need it, you can clean your PR by using `git reset --hard 503869e` then `git push -f`. It essentially tells git to forget that any other commit after the first one you wrote ever existed (it works because you don't have to keep a single lost commit inbetween).

PR #353, opened 2016-01-10
Expose isDragging to TargetMonitor

ywen:
The scenario where I need `isDragging` is described as following:

The drop targets, when no dragging, need to have certain CSS classes;
When deagging, if a specific target cannot be dropped it needs to have a
specific CSS class. And when it can it needs another CSS class.

The difficulty is that without `isDragging` the code cannot distinguish
between the first scenario (no dragging occurs) and the second (dragging
but cannot be dragged based on business rules)

### DIFF
--- a/src/createTargetMonitor.js
+++ b/src/createTargetMonitor.js
@@ -26,6 +26,10 @@ class TargetMonitor {
     }
   }
 
+  isDragging() {
+    return this.internalMonitor.isDragging();
+  }
+
   isOver(options) {
     return this.internalMonitor.isOverTarget(this.targetId, options);
   }


gaearon:
I merged this at first, but backed out. (I force-pushed the version without it.)

The reason is that we `isDragging()` on `DragSourceMonitor` does something different. It only returns `true` if *this particular source* is being dragged. It would be confusing to have `isDragging()` on `DropTargetMonitor` that works differently.

What you want should be already achievable by checking `Boolean(monitor.getItem())`. I agree it could be better but I don’t want to introduce inconsistencies at this point.

PR #355, opened 2016-01-15
Fixes #296 Cannot find valid target when multiple drop types specified

### LINKED ISSUE #296: Cannot find valid target when multiple drop types are specified
I found very interesting bug.

If I specify types for drop target in this way, it works as you'd expect:
```javascript
@DropTarget(['available', 'visible'], groupTarget, connect => ({
    connectDropTarget: connect.dropTarget(),
}))
export default class Group {
    ...
}
```

Since I need to set those types dynamically, I'd like to use function for that. According to docs, it should work - but it doesn't
```javascript
const dropTypes = props => props.accept;

@DropTarget(dropTypes, groupTarget, connect => ({
    connectDropTarget: connect.dropTarget(),
}))
export default class Group {
    ...
}
```

Btw, it works if I return string from function.

mahlero:
Fixes the issue from #296 where drop types are compared by reference only. If both drop types are arrays, an additional array comparison is performed.

### DIFF
--- a/src/decorateHandler.js
+++ b/src/decorateHandler.js
@@ -2,6 +2,7 @@ import React, { Component, PropTypes } from 'react';
 import { Disposable, CompositeDisposable, SerialDisposable } from 'disposables';
 import shallowEqual from './utils/shallowEqual';
 import shallowEqualScalar from './utils/shallowEqualScalar';
+import isEqual from 'lodash/lang/isEqual';
 import isPlainObject from 'lodash/lang/isPlainObject';
 import invariant from 'invariant';
 import bindConnector from './bindConnector';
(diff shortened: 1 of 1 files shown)


gaearon:
This fix happens to work, but this is not the underlying issue.
Description in https://github.com/gaearon/react-dnd/issues/296#issuecomment-165277413 is incorrect:

>Although the types are not changed occurs initialization below and takes some time.

JavaScript is single threaded (from user’s perspective) so a race condition is not the problem here. The problem is in DnD Core [removing handlers immediately](https://github.com/gaearon/dnd-core/blob/5569d80839ed429bea8ef2545c8b6cb4df57d58e/src/HandlerRegistry.js#L144-L146) which can cause problems during nested updates: for example, a `hover()` action causing `removeTarget()`.

The solution would be to change `HandlerRegistry` in `dnd-core` to delay deleting old handlers until the next tick.

gaearon:
Fixed in dnd-core via f1893a64965d2a294611d4d0449e5250a58308e0.

Issue #368, opened 2016-02-02
DragSource with "Stateless functional components" fails without error or warning


jackielii:
I'm not sure if this is expected behaviour. If so, some documentation is due?

Reproduce PR: #369

gaearon:
I don’t have the time to look into your PR in detail but functional components work fine on my end.
For example, I can replace

```js
@DragSource(ItemTypes.BOX, boxSource, (connect, monitor) => ({
  connectDragSource: connect.dragSource(),
  isDragging: monitor.isDragging()
}))
export default class Box extends Component {
  static propTypes = {
    connectDragSource: PropTypes.func.isRequired,
    isDragging: PropTypes.bool.isRequired,
    name: PropTypes.string.isRequired
  };

  render() {
    const { isDragging, connectDragSource } = this.props;
    const { name } = this.props;
    const opacity = isDragging ? 0.4 : 1;

    return (
      connectDragSource(
        <div style={{ ...style, opacity }}>
          {name}
        </div>
      )
    );
  }
}
```

with

```js
function Box({ connectDragSource, isDragging, name }) {
  const opacity = isDragging ? 0.4 : 1;

  return (
    connectDragSource(
      <div style={{ ...style, opacity }}>
        {name}
      </div>
    )
  );
}

export default DragSource(ItemTypes.BOX, boxSource, (connect, monitor) => ({
  connectDragSource: connect.dragSource(),
  isDragging: monitor.isDragging()
}))(Box);
```

in `01 Dustbin/Single Target/Box.js`, and it works fine.

PR #369, opened 2016-02-02
Issue #368 reproducible example

jackielii:
This is only for re-creating the issue, not for merging


gaearon:
Please see https://github.com/gaearon/react-dnd/issues/368#issuecomment-183782133.

PR #370, opened 2016-02-05
fix(docs): update examples to use findDOMNode from react-dom

Apercu:
Some people tend to be quite confused about this since the bump to React 0.14, this may help a bit!


gaearon:
Oh, good catch. Thanks.

PR #403, opened 2016-03-07
Remove the info about the dist folder

daviesgeek:
Since it was removed as of v2.0.0: https://github.com/gaearon/react-dnd/releases/tag/v2.0.0, there's no reason for that info to still be there, misleading unsuspecting victims like myself…


billyjanitsch:
Thanks @daviesgeek! FYI it looks like this was actually *added* in 2.0.0 but [removed in 2.1.0](https://github.com/gaearon/react-dnd/releases/tag/v2.1.0) in favor of npmcdn.

daviesgeek:
Whoops, sorry I meant deprecated…good to know though, thanks :)

PR #433, opened 2016-04-08
Upgrade React to stable (15.0.1)

antoinerousseau:
Current issue:

```
$ npm i --save react@15

npm ERR! peerinvalid The package react@15.0.1 does not satisfy its siblings' 
peerDependencies requirements!
npm ERR! peerinvalid Peer react-dnd@2.1.4 wants react@^0.14.0 || ^15.0.0-0
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "lodash": "^4.2.0"
   },
   "peerDependencies": {
-    "react": "^0.14.0 || ^15.0.0-0"
+    "react": "^0.14.0 || ^15.0.1"
   },
   "devDependencies": {
     "animation-frame": "^0.2.4",
@@ -75,10 +75,10 @@
     "mocha": "^2.2.5",
     "null-loader": "^0.1.0",
     "postcss": "^4.0.2",
-    "react": "^15.0.0-rc.2",
+    "react": "^15.0.1",
     "react-dnd-html5-backend": "^2.1.2",
     "react-dnd-test-backend": "^1.0.2",
-    "react-dom": "^15.0.0-rc.2",
+    "react-dom": "^15.0.1",
     "react-hot-loader": "^1.2.3",
     "react-router": "~0.13.2",
     "request": "2.46.0",


billyjanitsch:
Thanks -- can you also update the peer dependency, instead of just the dev dependencies? That's the one that's causing you trouble.

FYI, you should consider using npm@3 instead of 2 -- peer dependency errors have been replaced by warnings, so they don't block your build.

antoinerousseau:
oops yeah sorry the peer dependency was my initial goal! fixed.

Thanks for the tip for NPM 3, didn't know!!

billyjanitsch:
Thanks! :)

gaearon:
Um, no, `0-0` was intentional and according to http://semver.npmjs.com/ should match both 15 RC and 15 stable.

antoinerousseau:
Fair enough, but then why do I get `The package react@15.0.1 does not satisfy its siblings' peerDependencies requirements!` followed by `Peer react-dnd@2.1.4 wants react@^0.14.0 || ^15.0.0-0`?

But then I just realized that I was executing NPM v2.14.12, and when upgrading to NPM 3 the issue disappeared... weird.

gaearon:
I explain the problem a little more here: https://github.com/reactjs/react-redux/pull/342#issuecomment-207647767
Basically npm@2.x has very unhelpful behavior of listing *all* packages that peer depend on something even when *just one* peer dependency actually fails. As soon as you fix the actually problematic dependency, the error will go away. But yeah, you can upgrade to npm@3.x as well.

For the record, I reverted this PR because it has an unfortunate side effect of forcing people to use >=15.0.1 whereas there is no reason why they shouldn’t be able to stay on 15RC if they want to.

billyjanitsch:
Sorry for the mistake, @gaearon! I remember wondering whether `0-0` would actually match non-RCs -- I took this issue as confirmation that it didn't, but I should have double-checked.

gaearon:
I get confused by this all the time so I suggest using semver.npmjs.com to double check.

Issue #432, opened 2016-04-08
DropTarget that accepts all types


simshaun:
I'm trying to globally implement window scrolling when any DragSource is dragged close to the window's edge.

Going off of http://gaearon.github.io/react-dnd/examples-sortable-stress.html, I'm making my entire App a DropTarget whose sole purpose is to handle scrolling via the `hover` specification method.

What I can't figure out is how to have that DropTarget accept _any_ type. I don't want to have to explicitly list/register every possible type at the App root. Is that currently possible? or am I going about it all wrong?

froatsnook:
I'm not an expert, but as far as I know, this isn't possible.  Even if you get the `DragDropManager`, you'll only find types in `manager.registry.types` for rendered components (the `DragSource` type can be a function, and this function isn't called until it's needed).  One thing you could do would be to use your own `DragSource` that keeps track of types.

```diff
// YourComponents.js
- import { DragSource } from "react-dnd";
+ import { DragSource } from "./DragSource";
```

```javascript
// DragSource.js
import { DragSource as ReactDnDDragSource } from "react-dnd";
var types = new Set();
export function DragSource(type, ...args) {
  types.add(type); // hopefully you're not using (props) => type
  return ReactDnDDragSource(type, ...args);
}

export function allTypes() {
  return Array.from(types);
}
```

simshaun:
I'm wondering if it would be better to modify `DropTarget` to accept a wildcard (`true`, `*`, something like that) for acceptable types.

froatsnook:
Yeah I think that would be the right solution!  `true` would make sense and not break anyone's code who happened to use "*" as their source/target ID :).

froatsnook:
Hey @simshaun,

I implemented this in dnd-core and it seems to be working.  Before merging it in I need to update the tests some more, and then work on react-dnd.

@gaearon: what do you think of this change?  (`true` acting as wildcard for DropTargets)

simshaun:
Cool!

gaearon:
Yep, `true` sounds good to me. We’ll need to document this if we add this.

froatsnook:
Cool, I'll finish this up then.

crispy-invention:
@simshaun since we have canDrop() would you please see if this issue is still valid?

simshaun:
Sorry, it's been so long I don't remember what codebase I was working in.

[...]

PR #440, opened 2016-04-14
Feature/432 wildcard drop target

froatsnook:
New feature for #432 

There's a hard dependency on gaearon/dnd-core#31 since I'm using the new `isValidSourceType` and `isValidTargetType`

### DIFF
--- a/docs/00 Quick Start/Overview.md
+++ b/docs/00 Quick Start/Overview.md
@@ -24,7 +24,7 @@ What is an item? An *item* is a plain JavaScript object *describing* what's bein
 
 What is a type, then? A *type* is a string (or a [symbol](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Symbol)) uniquely identifying *a whole class of items* in your application. In a Kanban board app, you might have a `'card'` type representing the draggable cards and a `'list'` type for the draggable lists of those cards. In Chess, you might only have a single `'piece'` type.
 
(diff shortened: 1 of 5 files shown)


darthtrevino:
I'm going to close this out for now, it may make sense to have a predicate function here at some point, but I think using rest/spread to splat an array out would suffice.

enterthekiwi:
is this going to be merged? id very much like to use this. have a dnd component that can be used in many contexts and shouldn't have to know exactly which types it can accept

darthtrevino:
Probably not as-is. The original PR that @froatsnook submitted was two separate PRs across projects.  Since we've restructured to a monorepo the PR would need to be rebuilt.  I'd like to think through this feature request a bit more.  I'm not sure a wildcard is the most friendly way to approach this. A predicate function may make sense e.g. `@DropTarget(e => true)`, and maybe implicitly globally allowing drops would make sense e.g. `@DropTarget // no args, accepts everything`

enterthekiwi:
okay cool, thank you. i agree w/ the predicate function, but there should definitely be some solution to this problem so that dTargets don't have to know every single type dSource they need to respond to in advance.. it really makes it harder to build reusable dnd components, and i see no advantage since you can simply check the dragged item type in the hover/drop methods before doing anything. will keep an eye out for it, thanks for the work you've done on this library 😄

crispy-invention:
@froatsnook this is an old PR. We have canDrop() which can be used for similar purpose. Can you close this PR?

PR #451, opened 2016-04-27
Fixed typo

pmoons:


calesce:
`memoized` is valid and likely what he intended, see https://en.wikipedia.org/wiki/Memoization. It's a common pattern in React and related libraries.

pmoons:
Ahh, Apple's dictionary failed me.  Thanks for this new tidbit of knowledge. 🍻

PR #465, opened 2016-05-28
Remove unused import from docs

inooid:
While working with the `DragLayer` documentation I found a small unused import in the example code. It's a very trivial change, but to make the docs perfect, I'd suggest taking it out 😁


kesne:
Thanks for this!

PR #470, opened 2016-06-09
Change tutorial knight position to match screenshot

jfbarnes:
In the tutorial docs the knight position applied by hard-coding the prop doesn't match the screenshot. The image shows a knight in position [7,4] (instead of [4,7] like the code). 

Have updated the code example.


diegofaria:
I got confused there too, thanks!

PR #484, opened 2016-07-02
Update Overview.md, grammar

spefley:


kesne:
Correct me if I'm wrong, but it seems like less would be the correct usage here? http://www.quickanddirtytips.com/education/grammar/less-versus-fewer

kesne:
Going to close this out for now. Feel free to comment if you feel that I've made a mistake.

PR #486, opened 2016-07-12
Update docs to explain component parameter being null when using stateless components

MartijnHols:
Just some docs additions to hopefully help other people resolve this issue quicker;

when using the [Simple Sortable](https://github.com/gaearon/react-dnd/blob/master/examples/04%20Sortable/Simple/Card.js) example as a basis for something only slightly similar, I encountered the problem that the `component` parameter was always `null` and therefore could not be used to determine mouse position relative to my container. Couldn't find a solution in the docs or an existing issue, but after reading through some unrelated issues I saw a mention of stateless components and when I compared that part of my code with the example I found that was the cause of my issues. This PR updates the docs to hopefully provide others with a shortcut.


rhys-vdw:
@MartijnHols was just looking to see if there is a solution for this issue. I'm assuming the solution is to use stateful components, right?

kesne:
@rhys-vdw Yeah, the way to work around this is to just create a stateful component. I'm planning on making this a little more clear in the future because I've gotten bitten by this in the past.

@MartijnHols Thanks for this quick note. Seems like a good addition to the docs!

vieiralucas:
@kesne I guess something needs to be done to land this?
I just got bitten by this 🐱 
https://gaearon.github.io/react-dnd/docs-drag-source.html

PR #496, opened 2016-07-29
Fix an example of the tutorial

yogodoshi:
awesome tutorial btw! :D


kesne:
Thanks for the contribution!

Issue #222, opened 2015-07-02
Move to Redux?


JustinTulloss:
Are there plans to move this library from Flummox to Redux? I personally would be more willing to add Redux as a dependency if this library (which I already depend on) used it.

gaearon:
I don't have plans at the moment, but eventually I'd do that.

JustinTulloss:
Sounds good!

gaearon:
Done in `react-dnd@1.1.6` / `dnd-core@1.2.0`: https://github.com/gaearon/dnd-core/pull/21.

ouchxp:
@gaearon I suppose the documentation should be updated as well.
https://gaearon.github.io/react-dnd/docs-overview.html
The latest official document still saying it's backed by Flux.

gaearon:
Well, technically Redux is still rather similar to Flux. But a PR is welcome.

PR #501, opened 2016-08-06
Solves Issue #222 (documenting move to Redux)

zargold:
I was scared when reading the documentation that I'd have to bloat my dependencies to flux and Redux 👍 but don't have to 🍒 !


kesne:
Thanks for doing this!

PR #507, opened 2016-08-17
CSS fixes for square

bngsudheer:
minWidth: '12vw' and height: '12vw' fixes the square. Without them, the square is not visible on Chrome 52 and Firefox 48.


darthtrevino:
This looks okay in the latest versions of Chrome and FF

PR #517, opened 2016-08-23
Replace lodash with lodash.<fn> deps

### LINKED ISSUE #516: React-dnd from NPM is quite large, let's drop lodash?
Hey there, we're using react-dnd from NPM and we're quite surprised by the size of the library. The whole `react-dnd` is taking 5.7M. 4.7M of that is only `lodash`. From lodash you're using only `flatten`, `shuffle`, `isArray` and `isPlainObject` functions. These are available separately from NPM taking only 100K.

I know there are separate minified versions of this library, however, would you be interested in a PR that replaces the whole lodash with only what you guys are using?

jakubzitny:
Close #516

### DIFF
--- a/examples/01 Dustbin/Stress Test/Container.js
+++ b/examples/01 Dustbin/Stress Test/Container.js
@@ -4,7 +4,7 @@ import HTML5Backend, { NativeTypes } from 'react-dnd-html5-backend';
 import Dustbin from './Dustbin';
 import Box from './Box';
 import ItemTypes from './ItemTypes';
-import shuffle from 'lodash/shuffle';
+import shuffle from 'lodash.shuffle';
 import update from 'react/lib/update';
 
 @DragDropContext(HTML5Backend)
--- a/package.json
+++ b/package.json
@@ -42,7 +42,6 @@
     "dnd-core": "^2.0.1",
     "hoist-non-react-statics": "^1.2.0",
     "invariant": "^2.1.0",
-    "lodash": "^4.2.0",
     "lodash.flatten": "^4.4.0",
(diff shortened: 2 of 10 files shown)


jakubzitny:
The CI task exitted correctly, however, there is still some problem with the build. Let me know if you're actually interested in this and I'll update it.

billyjanitsch:
I'm 👎 on this. While the Lodash install size on disk is large, react-dnd uses modular imports (e.g. `lodash/shuffle`) to pull in only the parts that it needs, which keeps the bundle small.

The proposed change would actually *increase* the bundle size, since as of Lodash 4.14.0, the individual packages are zero-dependency modules and therefore, unlike modular imports, do not share any code, despite much of it being duplicate (see the [changelog](https://github.com/lodash/lodash/wiki/Changelog#v4140)). Moreover, the proposed change prevents apps which use both Lodash and react-dnd from sharing Lodash code.

I'm all for decreasing install size, but not at the cost of bundle size.

FYI, the individual packages are [on their way out](https://github.com/lodash/lodash/wiki/Roadmap#v500-2017) in Lodash 5.0.0 in favor of modular imports, and there are plans to significantly decrease install size.

kesne:
The individual packages being deprecated is a pretty compelling reason for me not to bring this in. 

With 5.0 lodash is aiming to shave 50% off of install size, so maybe we could just rely on that solving people's install size woes?

Alternatively, we could look into other modules instead. (e.g. using native `Array.isArray`, `array-shuffle`, `is-plain-object`).

jakubzitny:
These are valid points.

I'm closing the PR then. 😉

PR #538, opened 2016-09-15
Replace ref String Attribute with ref Callback Attribute

NameFILIP:
Fixes https://github.com/gaearon/react-dnd/issues/521


kesne:
Are you sure that this fixes the issue? I agree that callback refs are certainly what we should be doing, but I'm pretty sure stateless components can't be given refs no matter what.

NameFILIP:
I tested it locally. I entered:
```
(child) => { 
  console.log('Component is mounted'); 
  return (this.child = child); 
}
```
It did print the message and no more warning from the https://github.com/gaearon/react-dnd/issues/521

kesne:
I also see the message, but `child` is null, so it's likely just masking the problem, not solving it.

NameFILIP:
I see what you're saying. That's a valid point.
I'm afraid if you want to get an instance of the DecoratedComponent you'll have to make it stateful.

The question is: Does `react-dnd` requires a component to be stateful? 

I'm using Redux to maintain the state of my application and most of my components are stateless including the parent `App.jsx` which is being decorated. I don't plan to use the `getDecoratedComponentInstance()` and I am just trying to keep warnings out of my console.

I can always change the component to a `class` and trick the linter, but is it really necessary?
Any ideas on the solution?

NameFILIP:
@kesne I found a similar commit to the `react-redux`: https://github.com/reactjs/react-redux/commit/2d3d0beade55477b3af65534ceb793db18b25705
I don't like the idea of adding a `withRef` option. But I can add an `invariant` to the `getDecoratedComponentInstance`:

```
getDecoratedComponentInstance() {
  invariant(this.child, 'In order to access an instance of the decorated component it can not be a stateless component');
  return this.child;
}
```
What do you think?

kesne:
I think that'd be a much better approach to solving this.

For what it's worth, I think it'd be okay with the `withRef` option as well. That'd probably be a little bit better for performance than always attaching refs if you don't care about getting access to the component instance.

NameFILIP:
The `withRef = false` approach is a breaking change. I can implement it, but I'm not sure if such performance improvement is significant enough to potentially break existing applications. Let me know what you think.

kesne:
This LGTM. This might be a breaking change if anyone depended on the previous behavior of the ref being null.

NameFILIP:
Anything else I should add to this PR?

mattkrick:
if it's gonna be a breaking change either way, my vote is for the `withRef` as that seems to be the norm in the ecosystem (and is _slightly_ more performant). Nothing wrong with a major version bump if it means progress!

mattkrick:
@DanielRuf any chance we could this on a version bump in npm?

DanielRuf:
@mattkrick I'm not a contributor or collaborator with write access ;-)

mattkrick:
awh shucks, good call, those new GH review approvals got me all confused. 🙈

NameFILIP:
Added a fix to the tests (`ReactTestUtils` dependency).

Created an alternative PR using a `withRef` option (https://github.com/gaearon/react-dnd/pull/594). I actually prefer the first PR (current), because it's less likely to break an existing functionality and is easier to understand. Also, `withRef` option doesn't solve the ref to a stateless component issue.

But the person with merge capability is the judge and will be responsible.

PR #543, opened 2016-09-21
Missed a spot of Flux

zargold:
Should be "built in Redux" rather than Flux
In my last PR missed this one spot.

- [x] correctly indicates the Redux is used rather than Flux in index.md


kesne:
Thanks again for this!

PR #548, opened 2016-10-01
Remove duplicated 'canDrag' of ALLOWED_SPEC_METHODS

agilgur5:


kesne:
Nice catch! thanks for this!

PR #574, opened 2016-11-10
Add ES Modules build

salzhrani:
Also bumped the dependencies.
the `.babelrc` file is copied from [redux](https://github.com/reactjs/redux/blob/master/.babelrc)


darthtrevino:
.babelrc and lib updates will occur with PR #630

PR #594, opened 2016-12-14
Add withRef option

NameFILIP:
An alternative solution to https://github.com/gaearon/react-dnd/issues/521 . Also, see PR https://github.com/gaearon/react-dnd/pull/538


darthtrevino:
#538 was the winner, closing this PR

PR #606, opened 2017-01-05
Adds require for Box component to Testing.md

developering:
Without a require/import for the Box variable in the first example, it was not obvious that Box was an example of a user's component. It took me several times of reading through the description and the code before I realized what was happening. Adding an import should make it clearer to future readers.


developering:
Thanks @darthtrevino, good call. I just made that update.

PR #614, opened 2017-01-17
Adding a React Component for establishing Context via React Trees

darthtrevino:
This will allow us to nest the DragDrop engine under an iframe. Note that the iframe `window` must be injected via context.

see https://github.com/gaearon/react-dnd/issues/241
see https://github.com/gaearon/dnd-core/pull/37

### DIFF
--- a/src/DragDropContext.js
+++ b/src/DragDropContext.js
@@ -8,8 +8,8 @@ const CHILD_CONTEXT_TYPES = {
   dragDropManager: PropTypes.object.isRequired
 };
 
-const createChildContext = (backend, window) => ({
-  dragDropManager: new DragDropManager(backend, window),
+const createChildContext = (backend, context) => ({
+  dragDropManager: new DragDropManager(backend, context),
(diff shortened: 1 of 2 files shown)


kesne:
It seems weird to have users put the window object on the react context, I'm not sure what that solves?

darthtrevino:
The javascript at this point that's assembling the DOM is executed in the context of the outer window global, but the drag-and-drop context within the iframe needs to be constructed with the inner window to establish proper eventing.

darthtrevino:
We could probably use a prop to inject the inner window instead, but it seems kind of natural to use a context variable to replace a global ¯\\_(ツ)_/¯

darthtrevino:
I think what will make sense is to do the following checks to determine what window instance to use:
1) Check the `window` prop on this component. If it has been explicitly set, that should take the highest precedent.
2) Check the `window` context value. If this is available, it should be used.
3) Use the standard `window` global.

PR #618, opened 2017-01-19
Fixed invisible square at tutorial

diegofaria:
If width and height are not specified, the squares are invisible.


diegofaria:
Sure, it's done!

PR #661, opened 2017-02-13
redux added to webpack externals

mozkarakoc:


darthtrevino:
Is the intent here just to prune the bundle size? The DragDropManager is using redux to create a store internally. This would add a requirement that clients would need to load redux into the global scope first before using the prebuilt bundle. This would need a documentation update, and I'm not sure it would provide that much benefit.

If you _really_ need to prune your bundle down, I would avoid the prebuilt libs and generate your own bundle with webpack.

Issue #657, opened 2017-02-10
Support native html5 dropEffect


hpurmann:
Since this is a monorepo now, it probably makes sense to raise awareness of my issue here.

Please see https://github.com/react-dnd/react-dnd-html5-backend/issues/58

darthtrevino:
@hpurmann Feel free to cut a PR on this, I'll give it a fair review. The documentation examples also act as a kind of test-bed for different kinds of scenarios. In your PR, could you include an example that does what you're trying to pull off?  Dnd-core is pretty well tested, the rest of the project is so-so, as much as possible, please try to write unit tests to capture your desired behavioral changes.

Issue #394, opened 2016-02-26
holding ctrl while drag/dropping


soccertwin03:
Is there a mechanism to determine if a certain key is being pressed on drop?  This would be used for something similar to windows explorer. (drag/drop to move, drag/drop/ctrl to copy).

gaearon:
This is not currently supported by https://github.com/gaearon/react-dnd-html5-backend although I could imagine this would be possible with a custom backend. It might be possible to add there but I haven’t given it much thought. If you decide to look at it there, [`dropEffect`](https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/dropEffect) would be related.

From what I remember, right now we completely disregard the keys and instead let you specify the drop effect as an option to `connectDragSource` as [demostrated here](https://github.com/gaearon/react-dnd/blob/ca227203f0e3896789fbc29141b75d91701c0c7d/examples/05%20Customize/Drop%20Effects/SourceBox.js#L40). You might actually achieve what you want by using `setState({ dropEffect })` on key down and key up, and using `this.state.dropEffect` in your `connectDragSource` call.

I’m closing because I don’t plan to work on this, but hopefully these are enough pointers to get started.

gaearon:
According to https://github.com/gaearon/react-dnd-html5-backend/issues/23, the browser won’t trigger keydown and keyup during dragging so my earlier suggestion wouldn’t work.

I suggest you to track https://github.com/gaearon/react-dnd-html5-backend/issues/23. I don’t have plans to implement this but somebody else might.

hpurmann:
@soccertwin03 Maybe our proposal could help you: https://github.com/gaearon/react-dnd-html5-backend/issues/58

PR #675, opened 2017-03-01
Make alt drag work for drop sources that can take either copy or move drag operations

acusti:
Browsers treat users holding down the alt-key before or during a drag operation as a special case where if the `dropEffect` is `move`, the drop event will not fire. This PR makes react-dnd default the `dropEffect` to `copy` when the alt-key is pressed, which makes it so that by default, drop events still fire with or without holding alt, and passes its `dropEffect` to `dropResult`.

This addresses:

- https://github.com/react-dnd/react-dnd-html5-backend/issues/23
- https://github.com/react-dnd/react-dnd-html5-backend/issues/58
- #394
- #657

An easy way to see how this works is to pull down this branch and run react-dnd, then check the examples. All of the examples now work out of the box with users holding down alt, while the specific drop-effect example still works as before (no behavior changes there, because it sets a specific `dropEffect`, which overrides the default one).

Also, we are using a build of react-dnd-htm5-backend with this PR in place for our drag-and-drop implementation in [Brandcast](https://app.brandcast.io). If anyone wants to see it working in a real app, feel free to email me at andrew@brandcast.com and I will send you a signup invite link (currently a closed beta, but will be opened up to everybody in the next two weeks).

### DIFF
--- a/packages/react-dnd-html5-backend/src/HTML5Backend.js
+++ b/packages/react-dnd-html5-backend/src/HTML5Backend.js
@@ -27,6 +27,7 @@ export default class HTML5Backend {
     this.currentDragSourceNode = null;
     this.currentDragSourceNodeOffset = null;
     this.currentDragSourceNodeOffsetChanged = false;
+    this.altKeyPressed = false;
 
     this.getSourceClientOffset = this.getSourceClientOffset.bind(this);
     this.handleTopDragStart = this.handleTopDragStart.bind(this);
--- a/packages/react-dnd-html5-backend/src/HTML5Backend.js
+++ b/packages/react-dnd-html5-backend/src/HTML5Backend.js
@@ -144,7 +144,7 @@ export default class HTML5Backend {
     const sourceNodeOptions = this.sourceNodeOptions[sourceId];
 
     return defaults(sourceNodeOptions || {}, {
-      dropEffect: 'move',
+      dropEffect: this.altKeyPressed ? 'copy' : 'move',
     });
   }
 
@@ -397,6 +397,8 @@ export default class HTML5Backend {
       return;
     }
 
+    this.altKeyPressed = e.altKey;
+
     if (!isFirefox()) {
       // Don't emit hover in `dragenter` on Firefox due to an edge case.
       // If the target changes position as the result of `dragenter`, Firefox
@@ -437,6 +439,8 @@ export default class HTML5Backend {
       e.dataTransfer.dropEffect = 'none';
       return;
     }
+
+    this.altKeyPressed = e.altKey;
 
     this.actions.hover(dragOverTargetIds, {
       clientOffset: getEventClientOffset(e),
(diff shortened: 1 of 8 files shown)


darthtrevino:
@acusti This is looking pretty good; would you be willing to add a case in the docs for using alt when drag-dropping?

t1mmen:
@acusti Please contact me when this gets merged/released to claim you bounty, ref https://github.com/react-dnd/react-dnd-html5-backend/issues/23

packages/react-dnd-html5-backend/src/HTML5Backend.js
```diff
@@ -397,6 +397,8 @@ export default class HTML5Backend {
       return;
     }
 
+    this.altKeyPressed = e.altKey;
```
darthtrevino: We're setting `this.altKeyPressed` twice, is this necessary?

acusti: I’m not sure. This instance is for the `dragenter` handler, while the other is for `dragover`. From looking at http://stackoverflow.com/questions/32676890/want-to-know-the-difference-of-dragover-and-dragenter-events-of-html5-drag-and-d and https://social.msdn.microsoft.com/Forums/windows/en-US/be07bcde-6c99-4d68-824e-8ad93e820fe1/dragenter-vs-dragover?forum=winforms (ignoring the “Answer” and instead seeing the first reply under “All Replies”), it seems possible that dragenter could get invoked without dragover being called (on immediately entering a drop target), hence why I decided to just set the boolean in both handlers (in case the user dropped before a `dragover` event could be called).

acusti:
Thanks for reviewing, @darthtrevino! I’d be happy to add docs or similar for this behavior. Were you imagining a mention of the behavior and use cases in the docs, or a new example illustrating copy vs move in the examples/ folder?

darthtrevino:
A new example would be helpful for me, that way I can start the doc server and verify the work you've done

acusti:
@darthtrevino Checking back in on this. Regarding https://github.com/react-dnd/react-dnd/pull/675#discussion_r104500641, do you think my response makes sense for setting `this.altKeyPressed` both on `dragenter` and `dragover`? If you’d prefer, I’d be happy to update it to only set on `dragover`.

Also, have you had a chance to try the example I added? I hope it makes sense. Let me know if there’s anything else you’d like to see here. Thanks!

darthtrevino:
Thanks for the ping, @acusti, we'll get this one worked out today

darthtrevino:
The docs look good on my end; can you add `this.altKeyPressed = false`, to the constructor, then I'll merge this in

alvinhui:
This is an important feature, when will it be released?

darthtrevino:
It was released today, v2.3.0

cdauth:
It would be really nice if this supported other modifier keys as well. Using Alt seems to be Apple-specific behaviour. On Linux, Alt+Drag is typically not even possible because it moves the current window around.

PR #676, opened 2017-03-02
Update README.md

chen2009277025:
add Chinese document link。I translate it in chinese and add link


chen2009277025:
add Chinese document link。I translate it in chinese and add link

darthtrevino:
I'm going to close this for the time being. It looks like we're going to try and have better internationalization support in the documentation natively, and your translations may be useful then.

Issue #33, opened 2015-01-19
Demo does not work on mobile


winkler1:
http://gaearon.github.io/react-dnd/#/dustbin-interesting

... need to initialize React for touch?

gaearon:
It doesn't support touch (at least not yet).
The reason is we're basing on HTML5 Drag and Drop API and no touch browsers implement it.

This is not a *hard* limitation of this library. I do plan an optional touch backend in the future (see #16 and https://github.com/gaearon/react-dnd/issues/16#issuecomment-63743944 for groundlaying work), but since it's not a priority for my main job at the moment, I'm not actively developing that.

If you like the API and feel like doing some work to make it happen, ping me and I can help you get up to speed and give some hints as to how this can be implemented.

gaearon:
Closing this as a duplicate of #16. Feel free to chime in there!

winkler1:
Thanks. Will keep in mind if I work on a dnd project!

PR #695, opened 2017-03-13
Initializing class state in the Backend Constructor

darthtrevino:
The HTML5Backend initializes some arrays and local variables as methods are invoked on it. There are some unlikely edge cases where these pieces of state are being used prior to correct initialization. This commit explicitly sets a default initial state in the HTML5Backend constructor for those pieces of state.

See https://github.com/react-dnd/react-dnd-html5-backend/pull/33
See https://github.com/react-dnd/react-dnd-html5-backend/issues/32


darthtrevino:
@theTechie, does this look alright to you?

acusti:
Nice, I think this makes sense as an explanation for #33.

PR #711, opened 2017-03-26
fix: delete examples duplicate content

alvinhui:


darthtrevino:
Yikes, Thanks @alvinhui

PR #713, opened 2017-03-27
Fix intermittent problems with dropping a native file in FF

### LINKED ISSUE #539: File drop in Firefox doesn't always work
While working on a project it was noted that the native file drop didn't always work in Firefox. I didn't see any error or whatsoever but I did see the file drop was not handled by the function I had made therefore. While trying to create a testcase I made https://jsfiddle.net/r2mudx7a/ and there quite often the following error is thrown:

Error: Cannot call hover while not dragging.

When trying the fiddle please drag on a text file on the dashed square. In my latest test run I got the error nine times before it finally worked (the file was read and outputted to the console).

image

gcorne:
Fixes #539

When dropping a native file in FF, under some conditions the `mousemove` event was firing when the file was dropped before the `drop` event fired. If the drop operation would fail.

To fix, the event handler is delayed so that the end drag will always run after the drop handler. This commit also switches to using the `mouseover` event so that the drag state is reset sooner if the original FF dragleave bug is hit, and the commit resets the `EnterLeaveCounter` when ending the drag operation via the `mouseover` event handler because the FF `dragleave` bug results in `this.entered` not being empty.

### DIFF
--- a/packages/react-dnd-html5-backend/src/HTML5Backend.js
+++ b/packages/react-dnd-html5-backend/src/HTML5Backend.js
@@ -199,7 +199,7 @@ export default class HTML5Backend {
     this.currentNativeHandle = this.registry.addSource(type, this.currentNativeSource);
     this.actions.beginDrag([this.currentNativeHandle]);
 
-    // On Firefox, if mousemove fires, the drag is over but browser failed to tell us.
+    // On Firefox, if mouseover fires, the drag is over but browser failed to tell us.
     // See https://bugzilla.mozilla.org/show_bug.cgi?id=656164
     // This is not true for other browsers.
     if (isFirefox()) {
--- a/packages/react-dnd-html5-backend/src/HTML5Backend.js
+++ b/packages/react-dnd-html5-backend/src/HTML5Backend.js
@@ -43,6 +43,7 @@ export default class HTML5Backend {
     this.handleSelectStart = this.handleSelectStart.bind(this);
     this.endDragIfSourceWasRemovedFromDOM = this.endDragIfSourceWasRemovedFromDOM.bind(this);
     this.endDragNativeItem = this.endDragNativeItem.bind(this);
+    this.asyncEndDragNativeItem = this.asyncEndDragNativeItem.bind(this);
   }
 
   get window() {
@@ -74,6 +75,7 @@ export default class HTML5Backend {
     this.window.__isReactDndBackendSetUp = false; // eslint-disable-line no-underscore-dangle
     this.removeEventListeners(this.window);
     this.clearCurrentDragSourceNode();
+    clearTimeout(this.asyncEndDragTimeout);
   }
 
   addEventListeners(target) {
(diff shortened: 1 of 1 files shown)


gcorne:
To test the scenario where the `dragleave` FF bug is hit:

1. Resize and scroll window so that the drop target overlaps with the top of the browser window.
2. Drag file over the drop area and then leave the drop target by dragging out through the top of the browser window through the address bar. This will cause the hover state to be stuck until moving the mouse back over the window.

For more info on the FF bug see https://bugzilla.mozilla.org/show_bug.cgi?id=656164

packages/react-dnd-html5-backend/src/HTML5Backend.js
```diff
@@ -197,10 +199,19 @@ export default class HTML5Backend {
     this.currentNativeHandle = this.registry.addSource(type, this.currentNativeSource);
     this.actions.beginDrag([this.currentNativeHandle]);
 
-    // On Firefox, if mousemove fires, the drag is over but browser failed to tell us.
+    // On Firefox, if mouseover fires, the drag is over but browser failed to tell us.
+    // See https://bugzilla.mozilla.org/show_bug.cgi?id=656164
     // This is not true for other browsers.
     if (isFirefox()) {
-      this.window.addEventListener('mousemove', this.endDragNativeItem, true);
+      this.window.addEventListener('mouseover', this.asyncEndDragNativeItem, true);
```
gcorne: The switch `mosueover` makes the UX a little better because the user doesn't have to click to reset the drag state when the FF bug is hit.

packages/react-dnd-html5-backend/src/HTML5Backend.js
```diff
@@ -197,10 +199,19 @@ export default class HTML5Backend {
...
+    }
+  }
+
+  asyncEndDragNativeItem() {
+    this.asyncEndDragTimeout = setTimeout(this.endDragNativeItem, 100);
```
darthtrevino: Is the 100ms timeout necessary? Would a 0ms timework here to put this in the async queue?

gcorne: That was what I originally tried, but in testing, I still hit `endDragNativeItem` before the `drop` handler somewhat frequently. Since the state cleanup is not tied to any user intent, the added delay doesn't degrade the UX.

darthtrevino: Hmm, I'm not super comfortable depending on a magic number for a timeout; I'm worried that may lead to some flaky experiences down the road. The alternative solution was to simply remove the FF-specific event listeners. Would that work for your use case?

evertbouw: If you remove the event listener and then drag a file over a dnd app to a new browser tab, the app gets stuck in dragging state. You can get out of this state by dragging another file over the app. I don't think this is a huge problem but it'll be up to you guys if this is acceptable.

Delaying the mousemove event handler by using requestAnimationFrame also works and doesn't have magic numbers.

darthtrevino: Let's go with using requestAnimationFrame then; @gcorne, can you make that change?

gcorne: I agree that it is ugly. TBH, I think that there are actually two FF bugs at play. Mouse events firing before `drop` and the original `dragleave` issue.

> Would that work for your use case?

That will fix the issue with native file drop not working, which is a bigger issue than the issue with `dragleave` when it comes to drag-n-drop uploading. What I am not sure about is whether or not the `dragleave` issue is substantial enough that a fix should be kept. Another situation where the drag state can get stuck is when another OS window is overlapping with a drop target in a FF window and the user drags out of the window by moving over the overlapping window. I also spent some time looking for alternative workarounds (both myself and via Google) that were more deterministic, but didn't find any.

gcorne: Switched over in b5e06be

andrewQwer:
I can reproduce this issue with this fix :sob:

PR #720, opened 2017-03-31
Added support for CTRL key to copy as well as ALT

alexmorleyfinch:
On Windows, the copy modifier key is CTRL rather than ALT. This PR allows either key to copy

### DIFF
--- a/packages/react-dnd-html5-backend/src/HTML5Backend.js
+++ b/packages/react-dnd-html5-backend/src/HTML5Backend.js
@@ -27,7 +27,7 @@ export default class HTML5Backend {
     this.currentDragSourceNode = null;
     this.currentDragSourceNodeOffset = null;
     this.currentDragSourceNodeOffsetChanged = false;
-    this.altKeyPressed = false;
+    this.copyKeyPressed = false;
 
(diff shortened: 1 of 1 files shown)


darthtrevino:
I'm not opposed to this, but it should only use CTRL if the platform is Windows, and ALT if on OSX or linux

gshilin:
There is no reliable way to determine OS from  browser: [https://stackoverflow.com/questions/11219582/how-to-detect-my-browser-version-and-operating-system-using-javascript#answer-29057922](url)

Seldaek:
`var is_OSX = !!navigator.platform.match(/(Mac|iPhone|iPod|iPad)/i);` is a fairly reliable way of detecting OSX + iOS devices. Pretty much everything else (Windows/Linux) is using CTRL key, so that seems like a reasonable way to do the check.

megaboich:
I like this pull request, but probably it is better to think about making it configurable through some props. So it could be configured which key is going to be used as modifier key. And here I do see some lack of design - if for example I do need more than one modifier. For example I want to move an object by default, make a "link" if "Alt" is pressed and "copy" if "Ctrl" is pressed.
Also I could not find much of official documentation for this "dropEffect", all I could find is an example and it is not totally clear.

PR #729, opened 2017-04-08
Migrated deprecated React.PropTypes in example + same for React.creat…

Andarist:
…eClass in docs

I think we should migrate docs to the es6 `class` syntax and stop using `React.createClass`, but that would be another step. For now just created this PR, so people don't end up with deprecation warnings on React v15.5.

If you feel the same about `class`, I'll create an issue about it, so the matter won't get forgotten and neglected.

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,6 +73,7 @@
     "npm-run-all": "^4.0.1",
     "null-loader": "^0.1.0",
     "postcss": "^5.2.12",
+    "prop-types": "^15.5.4",
     "react": "^15.4.2",
     "react-addons-test-utils": "^15.4.2",
     "react-dom": "^15.4.2",
--- a/docs/00 Quick Start/FAQ.md
+++ b/docs/00 Quick Start/FAQ.md
@@ -16,9 +16,10 @@ You can see that in every example on the website that uses it, such as:
 
 -------------------
 ```js
+var createReactClass = require('create-react-class');
 var DragSource = require('react-dnd').DragSource;
 
-var MyComponent = React.createClass({
+var MyComponent = createReactClass({
   /* ... */
 });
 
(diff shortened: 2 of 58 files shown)


package.json
```diff
@@ -73,6 +73,7 @@
     "npm-run-all": "^4.0.1",
     "null-loader": "^0.1.0",
     "postcss": "^5.2.12",
+    "prop-types": "^15.5.4",
```
theTechie: did you miss adding `create-react-class` ?

theTechie: Should we be using `package.json` anymore since we are using yarn? @darthtrevino

Andarist: `React.createClass` was only mentioned in the docs, its not anywhere in the source code. So it's left out of dependencies.

Andarist: i think its important to keep `package.json` at least for npm users, npm must resolve dependencies based on this

Andarist:
@theTechie ive updated `yarn.lock`, but this repo is a multi package respository and im not sure if ive done it right as ive updated only the top level `package.json`/`yarn.lock`

package.json
rsolomon: Heads up, `prop-types` should be listed in `dependencies`, not `peerDependencies` or `devDependencies`. See gaearon's comment [here](https://github.com/callemall/material-ui/pull/6577). Also note that the build is failing with this message:

```
/home/travis/build/react-dnd/react-dnd/packages/react-dnd/src/decorateHandler.js
  2:1  error  'prop-types' should be listed in the project's dependencies. Run 'npm i -S prop-types' to add it  import/no-extraneous-dependencies
```

rsolomon: More specifically, it should be listed as a dependency in `react-dnd/[package.json|yarn.lock]` for the package usage itself, and the top-level as `devDependency`, because it's only used for examples and docs.

Someone else correct me if I'm wrong :)

Andarist: sounds reasonable, adjusted those

darthtrevino: @theTechie Yes, yarn works in tandem with package.json

darthtrevino:
Thanks for the PR, @Andarist!

Issue #761, opened 2017-05-05
Incorrect non-image drag offsets in Firefox on Retina display


pazams:
Steps to reproduce:
1) Use FF with retina display (`Mozilla/5.0 (Macintosh; Intel Mac OS X 10.12; rv:52.0) Gecko/20100101 Firefox/52.0`)
2) Load http://react-dnd.github.io/react-dnd/examples-dustbin-single-target.html
3) Drag any of the boxes (Glass , Banana, Paper) from the bottom right.

Actual Result:
Preview offset is off.

This might be similar to https://github.com/react-dnd/react-dnd/issues/4

PR #762, opened 2017-05-05
Fix cross-browser discrepancies for non-images

pazams:
Fixes https://github.com/react-dnd/react-dnd/issues/761


pazams:
Thanks for merging @darthtrevino. I've seen this is part of v2.4.1, is that version planned to be published to npm?

darthtrevino:
Whoops, I thought I did. I'll push out the release now

darthtrevino:
@pazams It looks like it did publish, but only the react-dnd-html5-backend - lerna must only publish individually changed repos on patch-level commits

Issue #775, opened 2017-05-22
Preview image, force offset via options


pschupp01:
Hello,

For the react-dnd-HTML5-backend, I would like to be able to force the vertical and horizontal offsets of the dragPreview via the options.
It would allow me to always have the same offset for my custom dragPreview images.

I can do a PR updating the minimal part if you want.

pschupp01:
I have this PR ready #776

PR #776, opened 2017-05-22
Pass offsetX and offsetY as createDragPreview options to force offsets

pschupp01:
Added two optionals fields (offsetX and offsetY) to createDragPreview options.
Those fields allow to force specific offsets values.

Linked to #775

### DIFF
--- a/packages/react-dnd-html5-backend/src/HTML5Backend.js
+++ b/packages/react-dnd-html5-backend/src/HTML5Backend.js
@@ -322,7 +322,9 @@ export default class HTML5Backend {
           clientOffset,
           anchorPoint,
         );
-        dataTransfer.setDragImage(dragPreview, offsetX || dragPreviewOffset.x, offsetY || dragPreviewOffset.y);
+        const forceOffsetX = offsetX === 0 || offsetX;
+        const forceOffsetY = offsetY === 0 || offsetY;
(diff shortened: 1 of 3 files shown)


bjrmatos:
any news on this? this looks very useful and i'm currently needing this, is there anything that i can do to help to land this feature?

pschupp01:
Still no news. I had to fork the repo and use my version.

pschupp01:
@darthtrevino Maybe this PR can be merge?

packages/react-dnd-html5-backend/src/HTML5Backend.js
```diff
@@ -314,15 +314,21 @@ export default class HTML5Backend {
         const sourceId = this.monitor.getSourceId();
         const sourceNode = this.sourceNodes[sourceId];
         const dragPreview = this.sourcePreviewNodes[sourceId] || sourceNode;
-        const { anchorX, anchorY } = this.getCurrentSourcePreviewNodeOptions();
+        const { anchorX, anchorY, offsetX, offsetY } = this.getCurrentSourcePreviewNodeOptions();
```
darthtrevino: I think we should pass in the the offsetX and offsetY as an additional argument to `getDragPreviewOffset` 
(offsetPoint) instead of doing the computation here.

pschupp01: Ok, i did the changes, it is now pass as an additional argument to `getDragPreviewOffset`.

I choosed to not optimize the method and even if offsets are forced the function still calculate the interpolated offset.

I didn't want to take any risk to break this part.

Issue #789, opened 2017-06-08
React DnD not work on Chrome 59 


hunterbmt:
After upgrading to Chrome 59,  react-dnd no longer works. After a bit investigating, it looks like Chrome no longer triggers "dragend" event which leads to react-dnd misunderstand when I try to drag the second item/time
Anyone experience the same issue?

theTechie:
Not the case; it work's perfectly fine. Try running Chrome without any extensions

hunterbmt:
@theTechie did you check on Mac on Windows ? On Mac it's working normally.

theTechie:
i checked on Mac. Is it a problem with a specific platform?

hunterbmt:
@theTechie for some reason, react-dnd don't trigger 'dragend' as normal. I tested on couple machine, one Mac has the same issue, all Windows has this issue. Any thought?

theTechie:
I tested on multiple platforms Windows 8.1 / Windows 7 on Chrome and also on IE 10 and IE 11. It works fine for me. 

@hunterbmt - Do you see any errors in the console ? could you share a GIF for some context?

hunterbmt:
@theTechie I tested on example site of react dnd, didn't have any issue. But in our application, Chrome 59 didn't trigger 'dragend' in some scenarios which works perfectly normal on Chrome 58, FF, Safari and even IE.
I fixed by manually trigger 'dragend' after 'drop' if the browser didn't trigger it after a while. 
I don't know if others fail into this issue as well, if that's the case, I could move the fix into react-dnd.

Looking at the changelog of Chrome 59 but didn't find anything, if you have any insight, I will appreciate the input a lot.

theTechie:
I think it could be something peculiar to your code; even i have used react-dnd in multiple instances and all of them work fine on Chrome 59.

Unfortunately, i am unable to tell you why this would be happening. Maybe, i can try and find something if you are ok sharing your code here.

awestroke:
I have the same problem in Chrome 59.0.3071.86 on Windows. It works fine in Firefox, and worker fine in earlier Chrome versions. package.json:
```json
{
  "dependencies": {
    "react-dnd": "2.1.4",
    "react-dnd-html5-backend": "2.1.2",
  }
}
```

awestroke:
Solved with this hack:

```js
const ModifiedBackend = (...args) => {
	const instance = new HTML5Backend(...args);

[...]

PR #801, opened 2017-06-19
Simplifying the flow. Observing 'mousemove' after 'dragstart' to trigger endDrag instead of depending on browser to trigger 'dragend'

hunterbmt:
Simplifying the flow, so can prevent the error to occurs when the browser fails to trigger 'dragend' or 'drop'. I.e https://github.com/react-dnd/react-dnd/issues/789
Now we depend only on 'mousemove' triggering after 'dragstart' to determine that the dragging is ended.

### DIFF
--- a/packages/react-dnd-html5-backend/src/HTML5Backend.js
+++ b/packages/react-dnd-html5-backend/src/HTML5Backend.js
@@ -32,7 +32,6 @@ export default class HTML5Backend {
     this.getSourceClientOffset = this.getSourceClientOffset.bind(this);
     this.handleTopDragStart = this.handleTopDragStart.bind(this);
     this.handleTopDragStartCapture = this.handleTopDragStartCapture.bind(this);
-    this.handleTopDragEndCapture = this.handleTopDragEndCapture.bind(this);
     this.handleTopDragEnter = this.handleTopDragEnter.bind(this);
     this.handleTopDragEnterCapture = this.handleTopDragEnterCapture.bind(this);
     this.handleTopDragLeaveCapture = this.handleTopDragLeaveCapture.bind(this);
@@ -41,7 +40,7 @@ export default class HTML5Backend {
     this.handleTopDrop = this.handleTopDrop.bind(this);
     this.handleTopDropCapture = this.handleTopDropCapture.bind(this);
     this.handleSelectStart = this.handleSelectStart.bind(this);
-    this.endDragIfSourceWasRemovedFromDOM = this.endDragIfSourceWasRemovedFromDOM.bind(this);
+    this.endDrag = this.endDrag.bind(this);
     this.endDragNativeItem = this.endDragNativeItem.bind(this);
     this.asyncEndDragNativeItem = this.asyncEndDragNativeItem.bind(this);
   }
@@ -83,7 +82,6 @@ export default class HTML5Backend {
   addEventListeners(target) {
     target.addEventListener('dragstart', this.handleTopDragStart);
     target.addEventListener('dragstart', this.handleTopDragStartCapture, true);
-    target.addEventListener('dragend', this.handleTopDragEndCapture, true);
     target.addEventListener('dragenter', this.handleTopDragEnter);
     target.addEventListener('dragenter', this.handleTopDragEnterCapture, true);
     target.addEventListener('dragleave', this.handleTopDragLeaveCapture, true);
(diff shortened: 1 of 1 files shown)


darthtrevino:
Hey @hunterbmt, thanks for the PR, sorry for not getting to it for a while. I'll give this a deeper look later tonight

hunterbmt:
@DanielKoehler thanks for taking care this one.

darthtrevino:
maui-in-moana-what-can-i-say-except-youre-welcome

packages/react-dnd-html5-backend/src/HTML5Backend.js
```diff
@@ -228,12 +225,7 @@ export default class HTML5Backend {
     this.currentNativeSource = null;
   }
 
-  endDragIfSourceWasRemovedFromDOM() {
-    const node = this.currentDragSourceNode;
-    if (document.body.contains(node)) {
```
bjrmatos: why this check is gone now? after seeing https://github.com/react-dnd/react-dnd/issues/789 i don't see any example that shows the problem.. the issue even says that the react-dnd example works normally in chrome 59

bjrmatos: please @darthtrevino consider revert this change, it is changing the behaviour of react-dnd.. 

- for example if i drag a component and drop it outside of a valid drop area and i don't move the mouse, the preview is hanging until i move the mouse (because this PR changed from using the native `dragend` event to `mousemove` event). 


before this PR (`endDrag` received inmediatly):

dragging before


with this PR (`endDrag` not received until i move the mouse):

dragging now


-before this PR if i start dragging something and press the "ESC" key the dragging is cancelled and then the `endDrag` method is called normally, with this PR that behaviour is gone

-before this PR the `endDrag` method was always called when the preview image (or ghost preview) actually returns to the original location where the dragging started, but now this is gone completly and the timing of the event is different. (with the additionally thing that we now need to mouse the mouse to get the `endDrag` method called)

since this is a HTML5 drag and drop implementation i think it is completly reasonable that we need to use the html5 drag and drop events, the browser handles a lot of cases automatically.. we are just getting in troubles by eliminating listening the `dragend` event.

if there is a problem with chrome 59 (like this [issue](https://github.com/react-dnd/react-dnd/issues/789) describes) we need to focus on reproduce the real problem instead of adding a workaround that changes the behaviour of everything. @hunterbmt can you please provide a minimal example where react-dnd fails with the `dragevent` in Chrome 59? i don't have any problem with Chrome 59, but i'm curious and willing to help to identify your real issue, if there is a problem it should be reproducible with a minimal example.

bjrmatos:
@darthtrevino i made some comments about this change, please let me know your feedback. thnks

bjrmatos:
and cc @hunterbmt just in case you don't get the notification for the mention in the review comment.

hunterbmt:
@bjrmatos I agreed with you about the behavior changes in the PR. I didn't think about all the cases when submitting the PR.
Since we alway check if we have currentSourceNode before trigger 'endDrag', I think having 'dragend' and 'drop' (for sourceNode removed from DOM cases) to trigger 'endDrag' beside 'mousemove' after 'dragstart' to trigger 'endDrag' won't hurt at all while covering all the case you mention. 
I cannot give you the example for Chrome 59 issues since it's a production code and it's quite a big application even though the logic at DnD component is quite straightforward. The code works normal on FF, IE, Safari and Chrome 58 but not Chrome 59 since it won't trigger 'dragend' event after 'drop' at all. In some cases, it won't even trigger 'drop' when I drag item into un-dropable area. But according to couple comments on the issues, I don't think that I face that issue alone.
I will create another PR to fix the issue without impact current behavior of react dnd. Sorry for a bad PR.
In the meantime, please help me to revert the PR first to avoid any accidentally release.

Regards,

darthtrevino:
It's not published yet, I wanted to test it some more before doing that.
I'll revert it

bjrmatos:
thnks @darthtrevino 



> I think having 'dragend' and 'drop' (for sourceNode removed from DOM cases) to trigger 'endDrag' beside 'mousemove' after 'dragstart' to trigger 'endDrag' won't hurt at all while covering all the case you mention.

let's see.. firing `endDrag` by listening `dragend` allows to handle a lot of functionality for free, so if you can make the change that you have in mind i'm willing to help to test that it does not change the current behaviour.


> I cannot give you the example for Chrome 59 issues since it's a production code and it's quite a big application even though the logic at DnD component is quite straightforward. The code works normal on FF, IE, Safari and Chrome 58 but not Chrome 59 since it won't trigger 'dragend' event after 'drop' at all. In some cases, it won't even trigger 'drop' when I drag item into un-dropable area. But according to couple comments on the issues, I don't think that I face that issue alone.

i understand that point but every issue needs at least a reproducible case, so maintainers or other people can help to verify the problem and think in the better solution, if there is no reproducible case the issue turns into something that maintainers or users of the library can not trust at 100% because there is no proof of the problem described. i understand that your codebase is big but trying to isolate the issue and removing some not relevant parts of your code to create a reproducible case should be not very hard and i think it is the only way to identify or verify the real issue so everyone can be happy.

PR #813, opened 2017-07-06
Check for presence of add/removeEventListener on target object

mrcthms:
I am currently using React-DND on our website, and client-side it works like a charm. However, with our server-side setup, it blows up. After doing a little digging, I found the code that was causing this was the `addEventListeners()` and `removeEventListeners()` methods in the `HTML5Backend.js` file.

The cause was that through our Express middleware, we assign some values to the `window` object server-side, primarily for things like translation strings, environment variables we need access to and so forth.

Because the check to fire these methods in HTML5Backend.js only checks whether `this.window` is undefined – and because technically the `window` object in our case is defined, but not an expected full client-side `window` object, we get the error when trying to run the `addEventListener()` and `removeEventListener()`.

So in this PR I propose a fix to simply return from the function early if the methods don't exist on the parent `target` object within these methods.

### DIFF
--- a/packages/react-dnd-html5-backend/src/HTML5Backend.js
+++ b/packages/react-dnd-html5-backend/src/HTML5Backend.js
@@ -81,6 +81,9 @@ export default class HTML5Backend {
   }
 
   addEventListeners(target) {
+    if (!target.addEventListener) {
+      return;
+    }
     target.addEventListener('dragstart', this.handleTopDragStart);
     target.addEventListener('dragstart', this.handleTopDragStartCapture, true);
     target.addEventListener('dragend', this.handleTopDragEndCapture, true);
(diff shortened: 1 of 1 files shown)


packages/react-dnd-html5-backend/src/HTML5Backend.js
```diff
@@ -81,6 +81,9 @@ export default class HTML5Backend {
   }
 
   addEventListeners(target) {
+    if (!target.addEventListener) {
```
darthtrevino: @mrcthms - Can you add a comment before these checks to indicate that these are SSR fixes, and include a link to this issue?

alexanderby:
I came here to find out what is SSR and I'm glad this is not a Soviet Socialistic Republic.

PR #848, opened 2017-08-16
update tutorial.md link

GeorgeWL:
https://react-dnd.github.io/react-hot-loader seems to not exist as a link anymore.

so changed to where I believe the link to now be: https://gaearon.github.io/react-hot-loader/


GeorgeWL:
@gaearon @react-dnd

Can we get a review on this please?

PR #892, opened 2017-09-27
Update some deprecated references

darthtrevino:
* There were some react/lib/x imports in the examples, replacing those with the recommended 'immutability-helper'.
* Updating react-hot-loader
* Adding a LICENSE file to the top-level project. Correcting SPDX license fields across the packages.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,28 @@
+BSD License
+
+Copyright (c) 2015, Facebook, Inc. All rights reserved.
+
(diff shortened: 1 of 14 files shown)


gaearon:
Why did this change license and add "(c) Facebook"? This doesn't make any sense. The project is, and will be MIT.

gaearon:
https://unpkg.com/react-dnd@0.1.0/LICENSE

darthtrevino:
I'm not sure when that was added, thanks for the clarification Dan

Issue #896, opened 2017-09-30
React 16 - Update: Warning: Stateless function components cannot be given refs.


jmdelcarmen:
Currently updated my project to React v16.0.0 from v15.6.2, and ran into this compatibility issue.

It appears that when the `DecoratedComponent` wrapped by `DragDropContext` and `DragSource` uses ref in React v16.0.0, this error is thrown. Error persists even if I refactor the component from stateless to class-based.

```js
warning.js:33 Warning: Stateless function components cannot be given refs. Attempts to access this ref will fail.

Check the render method of `DragDropContext(withRouter(Board))`.
    in withRouter(Board) (created by DragDropContext(withRouter(Board)))
    in DragDropContext(withRouter(Board)) (created by Cards)
...
```

package versions
* `react@^16.0.0`
* `react-dom@^16.0.0`
* `react-dnd@^2.5.3`
* `react-router@4.1.2`

haf:
If I have

```
 yarn list react-dnd
yarn list v1.3.2
warning Filtering by arguments is deprecated. Please use the pattern option instead.
└─ react-dnd@2.5.4
✨  Done in 0.93s.
```

And I'm still getting this error:

```
Warning: Stateless function components cannot be given refs. Attempts to access this ref will fail.

Check the render method of `DragDropContext(RendererRaw)`.
    in RendererRaw (created by DragDropContext(RendererRaw))
    in DragDropContext(RendererRaw) (created by Connect(DragDropContext(RendererRaw)))
    in Connect(DragDropContext(RendererRaw)) (created by App)
    in IntlProvider (created by Connect(IntlProvider))
    in Connect(IntlProvider) (created by App)
    in div (created by LoggingProvider)
    in LoggingProvider (created by Connect(LoggingProvider))
    in Connect(LoggingProvider) (created by App)
    in Provider (created by App)
    in App
```

For code:

```
// @flow
import { DragDropContext, DragDropContextProvider } from 'react-dnd';
// import HTML5Backend from 'react-dnd-html5-backend';
import MultiBackend from 'react-dnd-multi-backend';
import HTML5toTouch from 'react-dnd-multi-backend/lib/HTML5toTouch'; // or any other pipeline

export const Backend = MultiBackend(HTML5toTouch);
export { DragDropContextProvider as Provider }
export default (App: any) => DragDropContext(Backend)(App);
```

```
import {
  default as DragDropContext,
  Provider as DragDropContextProvider,
[...]

PR #898, opened 2017-10-05
do not attach refs on the stateless components

knowbody:
This fixes #896

PR #905, opened 2017-10-10
No refs on DragLayer

### LINKED ISSUE #904: React 16: Stateless function components cannot be given refs
I'm using the latest dnd v2.5.4 and still getting a console error for calls to `DragLayer`, which is set up in my app as a stateless function component. This is very similar to #896, which concerned `DragDropContext`

Log:

```
warning.js:33 Warning: Stateless function components cannot be given refs. Attempts to access this ref will fail.

Check the render method of `DragLayer(SitePageCustomDragLayer)`.
    in SitePageCustomDragLayer (created by DragLayer(SitePageCustomDragLayer))
    in DragLayer(SitePageCustomDragLayer) (created by SitePageContainer)
    in div (created by PanelBody)
    in PanelBody (created by SitePageContainer)
```

blord-fullscreen:
This fixes #904. Very similar to #896.

If the duplication of the isClassComponent function is a problem, tell me where to put helpers and I am happy to move things around.

### DIFF
--- a/packages/react-dnd/src/DragLayer.js
+++ b/packages/react-dnd/src/DragLayer.js
@@ -112,9 +112,11 @@ export default function DragLayer(collect, options = {}) {
 				return React.createElement(DecoratedComponent, {
 					...this.props,
 					...this.state,
-					ref: isClassComponent(DecoratedComponent) ?
-						child => { this.child = child } :
-						null,
+					ref: isClassComponent(DecoratedComponent)
+						? child => {
+								this.child = child
+							}
+						: null,
 				})
 			}
 		}
--- a/packages/react-dnd/src/DragLayer.js
+++ b/packages/react-dnd/src/DragLayer.js
@@ -7,6 +7,9 @@ import shallowEqual from './utils/shallowEqual'
 import shallowEqualScalar from './utils/shallowEqualScalar'
 import checkDecoratorArguments from './utils/checkDecoratorArguments'
 
+const isClassComponent = Comp =>
+	Boolean(Comp && Comp.prototype && typeof Comp.prototype.render === 'function')
+
 export default function DragLayer(collect, options = {}) {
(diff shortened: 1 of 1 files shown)


packages/react-dnd/src/DragLayer.js
```diff
@@ -106,15 +109,15 @@ export default function DragLayer(collect, options = {}) {
...
-				return (
-					<DecoratedComponent
-						{...this.props}
-						{...this.state}
-						ref={child => {
-							this.child = child
-						}}
-					/>
-				)
+				return React.createElement(DecoratedComponent, {
```
ranneyd: Why don't you use JSX?
```
<DecoratedComponent
    {...this.props}
    {...this.state}
    ref={isClassComponent(DecoratedComponent) ? child => this.child = child : null }
/>
```

ranneyd: Also, how sure are you that this works? Isn't the ref necessary?

blord-fullscreen: I am copying style from #898, which switched out JSX. I am trying to be consistent with the most recent changes in this vein. It appeared to be the preferred approach there, so I used it too! Sorry but this repo has no contrib guide, so I'm just trying to follow the current patterns. Switch it out if you like. But maybe do it in both places? I really just want to get this working with React 16.

ranneyd: For what's it's worth, I fixed it on my end by changing the actual component I was putting into `DragLayer` to a regular, non-stateless component. Since it was already a stateless component I just did it the `class` way and it worked. If you need it to work I'd recommend doing that.

blord-fullscreen: Thanks, I've actually already done that in my project. I just don't like using `class` for no good reason. (And it adds an Eslint error to my project, because I have a Component with just a render method.)

PR #921, opened 2017-11-16
Fix HTML5Backend intervene functioning of other file drag and drop library.

stkao05:
We notice an issue after integrating react-dnd into our app: Our existing drag-and-drop file upload function (which is not built on top react-dnd) failed to work. 

It seems like the event handling code `handleTopDragOver` in html5 backend will aggressively intervene the functioning of other DnD lib. 

```
if (canDrop) {
...
} else if (this.isDraggingNativeItem()) {
    // Don't show a nice cursor but still prevent default
    // "drop and blow away the whole document" action.
    e.preventDefault()
    e.dataTransfer.dropEffect = 'none'
} else if (this.checkIfCurrentDragSourceRectChanged()) {
...
}
```

I think it is ok to remove this part of the code, since it is not part this library's concern to prevent "blow way" scenario. 

(Please ignore `package.json` in the change file list 😅, I will remove this file if you think this PR makes sense)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,29 +1,9 @@
-[![npm version](https://img.shields.io/npm/v/react-dnd.svg?style=flat-square)](https://www.npmjs.com/package/react-dnd)
-[![npm downloads](https://img.shields.io/npm/dm/react-dnd.svg?style=flat-square)](https://www.npmjs.com/package/react-dnd)
-[![Build Status](https://travis-ci.org/react-dnd/react-dnd.svg?branch=master)](https://travis-ci.org/react-dnd/react-dnd)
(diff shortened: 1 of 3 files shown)


kepeterson:
Will this change stop react-dnd from eating all dragend events? If so I'm all for this change.

stkao05:
@kentcdodds Yes :) Glad to know others also experienced this issue.

kentcdodds:
Did you mention me by mistake? I think you may have meant to mention @kepeterson 😅

stkao05:
@darthtrevino 
The made the change in a separated branch and create a new PR. 
https://github.com/react-dnd/react-dnd/pull/1008

Will close this PR.

Issue #869, opened 2017-09-07
Not working in Firefox


mlukasch:
Hello,
reactdnd is not working anymore in the latest Firefox-Version 55.0.3
Using reactdnd version:
 "react-dnd": "^2.5.1",
 "react-dnd-html5-backend": "^2.5.1",

Does anyone knows the latest version of reactdnd where Firefox is working?

mlukasch:
React-dnd does work however in Firefox 55.0.2.

AKarmanov:
Doesn't work in 56.0.0 and beta 57.0.0 for me

Paddy-Hamilton:
not working in 56.0 either

alexfi:
Any ideas on this one?

bjrmatos:
can someone point to a repository with a sample case in which react-dnd is failing? i have an app that uses react-dnd a lot but i don't see this problem in firefox 56. 

also please make sure to specify the OS that you are using, i have seen that recently html5 Drag and Drop API is having some inconsistencies in different OS

lvpro:
I've observed issues with FF 56 on Linux, as well (didn't test on Windows or OSX).

http://react-dnd.github.io/react-dnd/examples-sortable-simple.html

Trying that simple example, an element will move at max 1 position and then no longer shift elements.

erwinverdonk:
For our case - DnD not working in FF, example https://react-dnd.github.io/react-dnd/examples-drag-around-naive.html - I did fix the issue with the pull request I submitted yesterday. Perhaps you could try and report back whether it solves your case as well.

bozskejkaja:
Still having issues on FF 57.0.1 on Ubuntu 16.04. I'm not able to drag & drop at all, seems like the drag event starts but cancels immediately afterwards. Works fine in chrome.

yankopetrov:
I also have problems with  FF and OSX 10.13.1

Paddy-Hamilton:
@yankopetrov 

I'm still experiencing the problems in Nightly, but it seems to break a whole lot more in the app than just RDandD. So not sure I trust that as an indication if FF is the real cause of the bug, and if they have fixed it.

Paddy-Hamilton:
@erwinverdonk , could you share your fix ? what was the PR?
cheers

yankopetrov:
@Paddy-Hamilton yes with FF Nightly still some bugs appears. Working with chrome and safari like a charm so should be FF issue. My case is exact the one in the reported FF issue.

[...]

PR #928, opened 2017-11-28
#869 - Fix issue which causes drag operation to not work on newer Fir…

erwinverdonk:
…efox versions and potential other browsers as well

Test this example on the newer Firefox versions and you will see that it doesnt perform the drag operation: http://react-dnd.github.io/react-dnd/examples-drag-around-naive.html

After this fix it does work, as the 'mouseover' event is now not triggered immediately because we now wait for the drag preview to be there before attaching the listener.


nerfologist:
This PR does indeed fix my issue with Firefox 57.0.1 (64-bit) on macOS High Sierra v. 10.12.6 (16G1114).

Issue description: upon starting to drag a draggable item, the drag&drop action would immediately cease (in a split second).

pianohacker:
This does not seem to fix the issue with Firefox 57.0.3 or 58.0.b14 on 64-bit Debian.

erwinverdonk:
How did you test the PR?

pianohacker:
@erwinverdonk I took the patch from your PR and applied it using patch-package.

I can dig a bit more (since I seem to be the exception), but is there anything I can look for to figure out if the patch didn't apply / why it isn't working? Debugging info I can turn on?

erwinverdonk:
@pianohacker I do not know `patch-package` that well, but my patch is done in `packages/react-dnd-html5-backend/src/HTML5Backend.js` while when the package is used it will actually load them from `packages/react-dnd-html5-backend/lib/HTML5Backend.js`. The `lib` folder is generated when the 'npm run build:lib' is called.

So maybe you applied the patch to the `src` folder through `patch-package`, while it should have been applied to the `lib` folder?

pianohacker:
@erwinverdonk Mrr, you were right. Kind of expected patch-package to rebuild the package, but anyway.

However, now that the patch IS in place, things aren't working much better. Dragging a DragSource between multiple DropTargets only works if I increase the time on the setTimeout call to at least 100 ms, and then only sometimes. It's possible I'm doing this wrong, but this all works flawlessly in Chrome.

erwinverdonk:
@pianohacker It could indeed be that in some situations or on some slower machines the drag source is still not rendered after an execution frame and the same bug still applies. The actual problem is the decision made to listen for the 'mousemove' event to determine drag has ended when the drag source is not on DOM anymore. 

@darthtrevino Can you enlighten us why this workaround (mousemove event to detect drag end) was added and whether it is still needed? :)

erwinverdonk:
Ok I have looked a bit better at my PR and tested a bit more thoroughly. I came to the conclusion that the fix is close, but its currently not it. So I will give it another go and hopefully it has more luck. PR will be updated soon.

erwinverdonk:
@pianohacker After some extensive testing I noticed, as you did too, that the timeout delay needs to be bigger. Therefore I increased the timeout to a second to make sure it will always add the event listener at the correct time.

darthtrevino:
Hey team, sorry for the delay on this. I'v been reticent to merge this one because of the setTimeout call. If all we need to do is wait an execution frame, why wouldn't a setTimeout with zero work? My concern here is that depending on a magic 1s timeout will be brittle in the long term.

packages/react-dnd-html5-backend/src/HTML5Backend.js
```diff
@@ -297,10 +298,21 @@ export default class HTML5Backend {
...
 		// so the browser didn't dispatch the dragend event.
-		this.window.addEventListener(
-			'mousemove',
-			this.endDragIfSourceWasRemovedFromDOM,
-			true,
+		//
+		// We need to wait an execution frame before we start listening for mousemove events.
+		// This is needed because the drag preview needs to be drawn or else it fires an 'mousemove' event
+		// immediately in some browsers.
+		this.mouseMoveTimeoutId = setTimeout(
```
darthtrevino: Can we rename this to `mouseMoveTimeoutTimer`?

packages/react-dnd-html5-backend/src/HTML5Backend.js
```diff
@@ -297,10 +298,21 @@ export default class HTML5Backend {
...
+			_ => {
+				_.mouseMoveTimeoutId = null
+				return _.window.addEventListener(
+					'mousemove',
+					_.endDragIfSourceWasRemovedFromDOM,
+					true,
+				)
+			},
+			1000,
```
darthtrevino: At the beginning of this file, add
```js
// A timeout of >0 is necessary because <insert justification here>.
// See:
//   * https://github.com/react-dnd/react-dnd/pull/928
//   * https://github.com/react-dnd/react-dnd/issues/869
const MOUSE_MOVE_TIMEOUT = 1000
```

Use this variable for the timeout value

nicksergeant:
I believe this code may have introduced a bug where we're attaching the `endDragIfSourceWasRemovedFromDOM` event listener to `mousemove` if you begin and end a drag before the 1 second timeout is reached.

This results in `endDragIfSourceWasRemovedFromDOM` being called on every subsequent `mousemove`, and `currentDragSourceNode` is `null` (since it was cleared out via the `dragend` event bound to `handleTopDragEndCapture` that occurred before the timeout was reached). Since `currentDragSourceNode` is `null`, it is not [cleared out](https://github.com/erwinverdonk/react-dnd/blob/ea1381b127415c7cf4387e9d9f71a9fa5e7e576c/packages/react-dnd-html5-backend/src/HTML5Backend.js#L301) until you begin and end a drag after the 1 second timeout.

volodymyrdn:
Hi there! Increasing the timeout to 1 sec, I encountered another problem: 

if the drop is done in less than 1 sec, endDrag method is not called: 
dnd2
Here you can see that the first drop is done after few seconds delay, and it works as expected. But the second drop is done in less than 1 second and the dropsource is not returned back to the list, it stays at the same place until the mouse move.

packages/react-dnd-html5-backend/src/HTML5Backend.js
volodymyrdn: Hi there!

PR #953, opened 2018-01-30
fix: eliminate unnecessary check in shallow equal

olegmedoed:


darthtrevino:
Thanks for pointing this out! We're going to use the `shallowequal` library instead of the custom implementations in this repo. (#1017)

Issue #925, opened 2017-11-26
Testing with Enzyme rather than React TestUtils


dwjohnston:
I'm getting on with writing some tests, following [this documentation](http://react-dnd.github.io/react-dnd/docs-testing.html) and also this issue https://github.com/react-dnd/react-dnd/issues/453

The problem I have is regarding the test rendering. 

I can get the example that the documentation gives working, wrapping the component in a DragDropContext and rendering it with `TestUtils.renderIntoDocument`. 

This will return a `DragDropContextContainer` object. 

However, if I try use the Enyzme `mount` method to render the object, it returns a `$$typeof: Symbol(react.element)` object - which won't let me get the manager. 

I would prefer to use Enzyme - as that's what I'm using for my tests elsewhere. 

Anyone got any help with this?

FreeMasen:
I just got this working, here is what it looks like. in typescript.

```ts
import { ThingIAmTesting } from '../file/where/it/lives';
describe('It is a test!', () => {
  let props;
  let component;
  const getComponent = () => {
    let OriginalComponent = (ThingIAmTesting as any).DecoratedComponent;
    let identity = el => el;
    if (!component) {
      component = Enzyme.shallow(
        <OriginalComponent {...props} connectDragSource={identity} />
       );
      return component
    }

    beforeEach (() => {
      props = {
         prop: 'stuff',
         }
      component = undefined;
     })
    it('should render', () => {
      let component = getComponent();
      expect(component).toMatchSnapshot();
    });
})
```

gilleswittenberg:
Ran into this same problem. I could, and expect you should be able to, fix it using the `wrapper.instance()` method. Also described here https://github.com/react-dnd/react-dnd/issues/488. 

e.g.
```javascript
const Context = wrapInTestContext(Component)
const wrapper = mount(<Context />)
const manager = wrapper.instance().getManager()
const backend = manager.getBackend()
```

Additionally, I also needed to get the instances of child components to call `getHandlerId()`.

If it helps, here is one of my test files: https://gist.github.com/gilleswittenberg/142fe49e1115b5a1936562b2284e1ab7

ericpoon:
Hey guys, I got an error saying **Invariant Violation: Expected sourceIds to be registered.** when I ran `backend.simulateBeginDrag([sourceId])`.

js
```
const wrapInTestContext = (DecoratedComponent) => {
    return DragDropContext(TestBackend)(
      class TestContextContainer extends React.Component {
[...]

PR #976, opened 2018-02-19
Improve documentation for testing

dsgkirkby:
This PR:
- Removes duplicate examples in the testing documentation
- Adds a new test to the Single Target example, testing a full drag-and-drop interaction
- Adds this same test to the testing documentation
- Adds a documentation section explaining what needs to be changed in the examples to work with Enzyme (addresses #925)

Note: I thought about adding a full-fledged test with Enzyme but I thought that adding Enzyme and an adapter as dependencies was a bit too heavyweight when a brief note in the docs should suffice. I'd be happy to update this with an Enzyme test if it's something that you think would be valuable.

### DIFF
--- a/packages/documentation/docs/00 Quick Start/Testing.md
+++ b/packages/documentation/docs/00 Quick Start/Testing.md
@@ -275,3 +275,19 @@ it('can be tested with the testing backend', () => {
 });
 ```
 -------------------
+
+### Testing with Enzyme
+
+[Enzyme](https://github.com/airbnb/enzyme) is a commonly-used tool for testing React components.
+To use it with react-dnd, you'll need to call `.instance()` on `mount`-ed nodes to access the react-dnd helper methods:
+```js
+var root = Enzyme.mount(<BoxContext name="test" />);
+
+// ...
+
+var backend = root.instance().getManager().getBackend();
+
+// ...
+
+backend.simulateHover([dustbin.instance().getHandlerId()]);
+```
--- a/packages/documentation/docs/00 Quick Start/Testing.md
+++ b/packages/documentation/docs/00 Quick Start/Testing.md
@@ -93,7 +93,7 @@ First, you need to install the test backend:
 npm install --save-dev react-dnd-test-backend
 ```
 
-Here is an example to get you started:
+Here are some examples to get you started:
 
 -------------------
 ```js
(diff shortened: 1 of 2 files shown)


darthtrevino:
@dsgkirkby Can you update your PR or allow us to update the source branch? There should be an option for that in the PR

dsgkirkby:
I have 'allow edits from maintainers' checked; is there something else I need to do?
I can also quickly fix this up tonight.

dsgkirkby:
I've updated the review from master.

packages/documentation/docs/00 Quick Start/Testing.md
```diff
@@ -158,86 +157,56 @@ import TestUtils from 'react-dom/test-utils';
...
- * Wraps a component into a DragDropContext that uses the TestBackend.
- */
-function wrapInTestContext(DecoratedComponent) {
-  return DragDropContext(TestBackend)(
-    class TestContextContainer extends Component {
-      render() {
-        return <DecoratedComponent {...this.props} />;
-      }
+it('can simulate a full drag and drop interaction', function() {
+  var DustbinWithBox = createReactClass({
```
darthtrevino: const

darthtrevino: Can you use a SFC here?

packages/documentation/docs/00 Quick Start/Testing.md
```diff
@@ -158,86 +157,56 @@ import TestUtils from 'react-dom/test-utils';
...
-  // Find the drag source ID and use it to simulate the dragging operation
-  const box = TestUtils.findRenderedComponentWithType(root, Box);
-  backend.simulateBeginDrag([box.getHandlerId()]);
-
-  // Verify that the div changed its opacity
-  div = TestUtils.findRenderedDOMComponentWithTag(root, 'div');
-  expect(div.style.opacity).toEqual(0.4);
-
-  // See other backend.simulate* methods for more!
+    var DustbinWithBoxContext = wrapInTestContext(DustbinWithBox);
```
darthtrevino: Use const

packages/documentation/docs/00 Quick Start/Testing.md
```diff
@@ -158,86 +157,56 @@ import TestUtils from 'react-dom/test-utils';
...
-  const BoxContext = wrapInTestContext(Box);
-  const root = TestUtils.renderIntoDocument(<BoxContext name='test' />);
+### Testing with Enzyme
 
-  // Obtain a reference to the backend
-  const backend = root.getManager().getBackend();
+[Enzyme](https://github.com/airbnb/enzyme) is a commonly-used tool for testing React components.
+To use it with react-dnd, you'll need to call `.instance()` on `mount`-ed nodes to access the react-dnd helper methods:
+```js
+var root = Enzyme.mount(<DustbinWithBoxContext />);
```
darthtrevino: const

packages/documentation/docs/00 Quick Start/Testing.md
```diff
@@ -158,86 +157,56 @@ import TestUtils from 'react-dom/test-utils';
...
-  }
-
-  return TestContextContainer;
-}
-
-it('can be tested with the testing backend', () => {
-  // Render with the test context that uses the test backend
-  const BoxContext = wrapInTestContext(Box);
-  const root = TestUtils.renderIntoDocument(<BoxContext name='test' />);
+### Testing with Enzyme
```
darthtrevino: This enzyme example is great, but I don't want to blow away the docs describing the standard tooling. Can you make these separate section instead?

dsgkirkby: Maybe it's changed since the original PR, but when I first made this change the examples were duplicates of each other?

dsgkirkby:
I've updated the review to not clobber the old examples, as they now look like they're examples of doing the same thing with different levels of language features. I've implemented the full drag-and-drop example just for the first section in ES5, as it seems unnecessarily repetitive to add it for each section.

PR #979, opened 2018-02-26
Pass props to DragLayer

casr:
I have found this useful for when the mouse x/y are too granular and therefore update the component much more frequently than required.

Example:
```js
render() {
  return (
    <CustomDragLayer gridWidth={3} gridHeight={3} />
  )
}
```
```js
const collect = (monitor, props) => {
  const clientOffset = monitor.getClientOffset()
  const { gridWidth, gridHeight } = props

  if (clientOffset != null) {
    return {
      gridX: Math.floor(clientOffset.x / gridWidth),
      gridY: Math.floor(clientOffset.y / gridHeight)
    }
  }

  return { gridX: null, gridY: null }
};
DragLayer(collect)(CustomDragLayer);
```


darthtrevino:
@casr Interesting, I like the idea! Can you write up an example of using custom drag layer props? In your example here I don't see the new props argument being used at all.

casr:
Ah I missed off a crucial line:
`const { gridWidth, gridHeight } = props`

I will amend the example above.

PR #988, opened 2018-03-21
Build Updates

darthtrevino:
* Use Yarn Workspaces to eliminate the `lerna bootstrap` step. 
* Include eslint-config-prettier to cull down some of our custom configuration to disable style-centric eslint rules.


wootencl:
@darthtrevino May already be aware but just a heads up it looks like this PR broke the `Browse the Source` links for the examples in the documents.

darthtrevino:
Natch, thanks for the heads up

PR #1017, opened 2018-05-01
Typescript Migration

darthtrevino:
The API for this project is complex enough to warrant a conversion to a type-system. This PR will migrate the react-dnd project to Typescript, while performing some minor structural and build updates along the way. The API should remain largely unchanged but well typed.

**Comments and reviews are welcome**

**Package Migrations**
- [x] Migrate dnd-core
- [x] Migrate react-dnd-test-backend
- [x] Migrate react-dnd
- [x] Migrate react-dnd-html5-backend

**Development/Tooling Updates**
- [x] Add wallaby configuration
- [x] Add ts-loader to webpack config
- [x] Upgrade to webpack 4
- [x] Upgrade eslint
- [x] Enable strict-mode rules
- [x] Add tslint for Typescript code
- [ ] Fill in `any` stopgap typings
- [ ] Minimize dynamic 'as' casting

**Structural Goals**
- [x] Replace 'shallowequal' implementations with npm dependency
- [x] Remove window and DOM dependencies from `react-dnd` to allow for non-browser scenarios eventually ( Note: **Breaking Change** w/ DragDropContext and DragDropContextProvider components)
- [x] (maybe) Update dnd-core actions to be FSA-compliant
- [ ] (maybe) Update dependency-injection so that constructors don't create dependencies

**QA**
- [x] Verify min vs nonmin bundles
- [x] Verify public surface area of key classes (e.g. HTML5Backend)


damiangreen:
Is the Simple Sorting example to be recreated in typescript as well?
Is there a version which doens't use ES7 decorators?

darthtrevino:
The documentation needs a bit of love TBH. Right now most examples have a switch between ES5, ES6, and "ES7" with decorators. I'd like to drop the ES5 portions and possibly add typescript as a category. I'm not sure if the Simple Sorting example in particular has a decorator-free option currently, but if you look at how the decorator examples work vs decorator-free versions, you can pick apart how to convert it.

Kadrian:
Could it be that this effectively dropped IE11 support, because `ES5` is not the target?

PR #1022, opened 2018-05-09
Add clarification to FAQ about Babel

edele:
Using `babel-react-optimize` preset may break react-dnd. It is particularly hard to diagnose if you only use the preset in production only, which seems reasonable. This little clarification in FAQ hopefully saves a lot of time for future users of react-dnd.


darthtrevino:
I realize it's been a year since you posted this, but I'm going to include a paraphrased version of this in the FAQ. Thanks for your contribution!

PR #1031, opened 2018-05-25
Improve Typings

darthtrevino:
This PR updates the typings in the project to more closely match DefinetlyTyped's type definitions. There are some minor differences, but we use the same naming strategy overall and correct typings in some corners.


darthtrevino:
Note to self: The sortable examples aren't working right now, get those fixed up

PR #1054, opened 2018-06-01
[HTML5 Backend] React should be a dev dependency

abettadapur:
HTML5 Backend listed React as a core dependency, which is incorrect as it is also listed as a peer dependency.

Instead, move this as a devDependency


abettadapur:
@darthtrevino Could you look at this small change?

darthtrevino:
Derp. Good catch

darthtrevino:
I'll cut a patch release after lunch

PR #681, opened 2017-03-04
html5 backend support dataTransfer.setData

bolasblack:
In my case I need to make todo item can be drag to native textfield of operating system and become some text like `- [ ] todo description`.

So I need call `dataTransfer.setData` in my component like :

```jsx
const operateDataTransfer = dataTransfer => {
  const completeMark = todo.get('completed') ? 'x' : ' '
  const markdown = `- [${completeMark}] ${todo.get('description')}`
  dataTransfer.setData('Text', markdown)
  dataTransfer.setData('text/plain', markdown)
  dataTransfer.setData('text/markdown', markdown)
  dataTransfer.setData('application/json', todo.toJS())
}

const dragSource = connectDragSource(
 <span className="handler">
   <i className="icon-handler" style={isSortable ? {} : {visibility: 'hidden'}}></i>
 </span>,
 { operateDataTransfer },
)
```

I'm not sure the code about `dataTransfer.setData` is a good solution, but I have no better idea :(

PR #1079, opened 2018-06-13
HTML5Backend: support passing data to dataTransfer.setData via DragSource beginDrag

fde31:
I guess this one is very similar to #681 in a sense that it enables a DragSource spec to pass additional Data to the backend that will be assigned to the underlying dataTransfer using `setData`. This PR is a somewhat initial, naive approach I guess but I'd be more than happy to start a discussion and do further work/refinements if this is a desired feature that could be useful for others.

The use-cases that I'm facing is that this way a browser hosted UI/App can interact with other applications and share data that goes beyond the borders of the current Document. In order to set the data one would include a `dataItems` props with `[mimeType]: "value"` pairs from within beginDrag of a DragSource spec.

As mentioned earlier the main difference is that this uses beginDrag instead of supplying the additional data as an option to connectDragSource as done in #681.

### DIFF
--- a/packages/react-dnd-html5-backend/src/HTML5Backend.ts
+++ b/packages/react-dnd-html5-backend/src/HTML5Backend.ts
@@ -431,8 +431,15 @@ export default class HTML5Backend implements Backend {
 			}
 
 			try {
-				// Firefox won't drag without setting data
-				dataTransfer.setData('application/json', {})
+				const { dataItems } = this.monitor.getItem()
+				if (dataItems && typeof dataItems === 'object') {
+					Object.keys(dataItems).forEach(k =>
+						dataTransfer.setData(k, dataItems[k]),
+					)
+				} else {
+					// Firefox won't drag without setting data
+					dataTransfer.setData('application/json', {})
+				}
 			} catch (err) {
 				// IE doesn't support MIME types in setData
 			}


fde31:
@darthtrevino makes total sense and yeah, that's why I put it up for discussion. Utilising `JSON.stringify` seems like a worthwhile approach especially given that the dataTransfer might mostly be used for cross-application data sharing. There are further possible adjustments in order to cover more of the spec and provide built-in practices.

I'm hereby thinking about making sure that:

* drag-types are lower-case
* `Text`, `text`, `text/unicode` map to `text/plain`
* `text/plain` is added as last/default item
* set `text/plain` when only `text/uri-list` is present to provide a sensible default
* allow `text/html` to be either a String or a DOM Node, in which case one could use innerHTML automatically (again maybe with a sensible `text/plain` default if not present)

I also wonder how much "sugar" react-dnd should provide for `text/uri-list` as the user could also just pass in an array and assuming correct concatenation using line-breaks.

I'm happy to work on this and update the PR but I thought it might be better to have the current contributors chip in their opinions in order to get a feel of how much built-in, "magic" functionality the package can/should offer. In general the items above can be helpful for lots of cases but I do wonder if they might get in the way, when custom functionality is desired due to an external application expecting a certain data type and/or format.

PR #1082, opened 2018-06-15
fix case where coords are null in beginDrag

### LINKED ISSUE #986: monitor.getClientOffset() return null in beginDrag
Hi,

I wonder if I can determine what beginDrag returns by { x, y } of the pointer. Like this:
```js
  beginDrag(props, monitor, component) {
    const canvas = component.context.canvas; 
    const pointer = monitor.getInitialClientOffset();
    return canvas.findTarget(pointer) || {};
  }
```
I just can't wrap things in the canvas as a drag source.

mattkrick:
This is a proposed solution to fix #986 

Once I get word that this will be merged, I'll 
- write tests
- clean up BEGIN_DRAG by removing stuff pertaining to coords
- make INIT_COORDS more dry

until then i'm just sharing this with the community in case they need a way to work around #986


packages/dnd-core/src/actions/dragDrop.ts
```diff
@@ -23,6 +24,47 @@ export default function createDragDropActions<Context>(
...
+      }: InitCoordsOptions
+    ): Action<InitCoordsPayload> | undefined {
+      const monitor = manager.getMonitor()
+      const registry = manager.getRegistry()
+      invariant(!monitor.isDragging(), 'Cannot call beginDrag while dragging.')
+
+      for (const s of sourceIds) {
+        invariant(registry.getSource(s), 'Expected sourceIds to be registered.')
+      }
+      let sourceId = null
```
darthtrevino: Put lines 41-47 into a helper function, named something like "getFirstSourceId"

jiholim:
https://flatuicolors.com/

PR #1086, opened 2018-06-18
Fix some reported bugs

### LINKED ISSUE #1080: endDrag not being called when dropping target into valid target
**Describe the bug**
According to the [documentation](http://react-dnd.github.io/react-dnd/docs-drag-source.html):

> For every beginDrag call, a corresponding endDrag call is guaranteed.

But using versions 4.0.5 of React DnD and DnD HTML5 backend, the endDrag event is only called when the DragSource is dropped in a place where it can't be dropped, e.g. outside the drop area

**To Reproduce**
```
const eventSource = {
    beginDrag(props) {
        console.log("Starting drag")
    },
    endDrag(props) {
        console.log("Ending drag")
    }
};
```

**Expected behavior**
There should be a print of "Ending drag" after each "Starting drag"

**Screenshots**
image

### LINKED ISSUE #1084: HTML5Backend should not have 'react' in peerDependencies either
I'm the author of [`angular-skyhook`](https://github.com/cormacrelf/angular-skyhook/), which invites users to install `react-dnd-html5-backend`. Installing it complains about a missing peer dependency `react@>= 16.3`.

As far as I can tell, the backend doesn't require React at all, and that is the reason why `angular-skyhook` exists. Could you leave that peer dependency for `react-dnd` to handle?

Tracking: https://github.com/cormacrelf/angular-skyhook/issues/1

darthtrevino:
Fix #1084 (remove react peerDep in html5 backend)
Fix #1080 (update the DragSourceSpec and DropTargetSpec interfaces to allow null components)

### DIFF
--- a/packages/react-dnd-html5-backend/package.json
+++ b/packages/react-dnd-html5-backend/package.json
@@ -37,8 +37,5 @@
 		"ts-loader": "^4.2.0",
 		"typescript": "^2.9.1",
 		"webpack": "^4.6.0"
-	},
-	"peerDependencies": {
-		"react": ">= 16.3"
 	}
 }
--- a/packages/react-dnd/src/createSourceFactory.ts
+++ b/packages/react-dnd/src/createSourceFactory.ts
@@ -51,7 +51,7 @@ export default function createSourceFactory<
 
 	class SourceImpl implements Source {
 		private props: P | undefined
-		private component: TargetComponent | undefined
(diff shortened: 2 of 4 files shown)


darthtrevino:
It looks like fixing #1080 will cause a major semver with typings. I'll fix #1084 in a separate pr and continue working on 1080

PR #1120, opened 2018-08-17
Add DATA_TRANSFER NativeType to access dataTransfer object

### LINKED ISSUE #712: HTML5 backend: access to DataTransfer.items when dropping files
This is needed to be able to drop directories and get their content (using `DataTransferItem#webkitGetAsEntry`, cf https://developers.google.com/web/updates/2012/07/Drag-and-drop-a-folder-onto-Chrome-now-available)

Currently `dataTransfer.files` is exposed, and by looking at the code (https://github.com/react-dnd/react-dnd/blob/master/packages/react-dnd-html5-backend/src/NativeDragSources.js#L16) I dont see how to expose multiple attributes.

I can give a try at implementing it if you tell me the preferred way to do so.

robertlong:
Fixes #712 without creating a breaking API change by adding `NativeTypes.DATA_TRANSFER`.

The data returned will be the [DragEvent.dataTransfer](https://developer.mozilla.org/en-US/docs/Web/API/DragEvent/dataTransfer) object.


renchap:
This is an interesting take on the problem, and solves it without breaking compatibility 👍

quentez:
Would love to see this merged.

We want to be able to count the number of files before they get dropped, and we can't do that without access to the underlying DataTransfer object.

darthtrevino:
Hey everyone,
I'm going to have to revert this change, as it interferes with existing functionality, namely how native file-drop works (the native file example is broken ATM).

The problem is in this function (https://github.com/react-dnd/react-dnd/blob/a403c42f8ed1e02c97633dc5116789fbfb9f49a3/packages/react-dnd-html5-backend/src/NativeDragSources.ts#L91-L100) The first NativeType that matches is used to extract the native type, and because this is at the top of the object definition, it's smashing everything below it.

I'd be glad to take a look at other approaches, but we can't compromise existing functionality.

Thanks!
Chris

PR #1149, opened 2018-10-29
Lib Upgrades (WIP)

darthtrevino:
- [x] Build Tooling: update husky, prettier, npm-run-all, lerna, jest
- [x] Webpack: update webpack, webpack-dev-server, webpack-cli, loaders
- [x] Boilerplate Tooling: update lodash, shallowequal, babel
- [x] add Commitlint Verification
- [x] remove @types/node dependency, since it was only used for ambient `require()` definition
- [x] upgrade React
- [x] upgrade Typescript
- [x] remove less, streamline gh-pages deployment


darthtrevino:
I think the docsite will eventually change over to being Gatsby or something, but this should drop the dependency count and simplifies the doc-site a bit.

PR #1157, opened 2018-11-07
Fix documentation links

TheTimBecker:


darthtrevino:
This will be fixed with #1158

Issue #1148, opened 2018-10-26
Hooks API Discussion


jacobp100:
Somebody was going to ask eventually! The new hooks API could possibly help here. I think most of the API can pretty much stay the same, with the HOC mapping directly into a hook.

I do wonder if we can replace `connectDragSource` and `connectDropTarget` with just passing in the value of `useRef`. It could definitely make stuff cleaner if that's possible!

darthtrevino:
I can't wait to start using hooks in this library. As soon as the types are baked for the alpha we can set up a migration branch

darthtrevino:
I'm playing around in a branch: experiment/hooks, just to think about how this API might look. The BoardSquare component is looking like this:

```jsx

const dropTarget = createDropTarget(ItemTypes.KNIGHT, {
	canDrop: (props: BoardSquareProps) => canMoveKnight(props.x, props.y),
	drop: (props: BoardSquareProps) => moveKnight(props.x, props.y),
})

const BoardSquare = ({ x, y, children }) => {
	const black = (x + y) % 2 === 1
	const [connectDropTarget, isOver, canDrop] = useDnd(
		dropTarget,
		connect => connect.dropTarget,
		(connect, monitor) => !!monitor.isOver,
		(connect, monitor) => !!monitor.canDrop,
	)

	return connectDropTarget(
		<div>
			<Square black={black}>{children}</Square>
			{isOver && !canDrop && <Overlay color={'red'} />}
			{!isOver && canDrop && <Overlay color={'yellow'} />}
			{isOver && canDrop && <Overlay color={'green'} />}
		</div>,
	)
}
```

darthtrevino:
So the idea here is that the createDragSource and createDropTarget would specify the item type and predicates about its behavior, and then the hook would wire this into the DnD backend and provide collected props.

jacobp100:
What branch are you working on? I'm wondering if we can remove `connectDropTarget` all together using refs. I'd love to see if I can get it working on your branch!

```js
const dropTarget = createDropTarget(ItemTypes.KNIGHT, {
	canDrop: (props: BoardSquareProps) => canMoveKnight(props.x, props.y),
	drop: (props: BoardSquareProps) => moveKnight(props.x, props.y),
})

const BoardSquare = ({ x, y, children }) => {
	const dropTargetElement = useRef(null);
	const black = (x + y) % 2 === 1
	const [isOver, canDrop] = useDnd(
		dropTargetElement,
		dropTarget,
[...]

PR #1173, opened 2018-11-17
Implement Hooks API

jacobp100:
We've had a bit of discussion over on #1148. Here's my attempt at exposing a hooks example!

You can see the hooks in use in the chessboard example.

* Adds `useDragSource` and `useDropTarget` return a `dragSourceMonitor` or `dropTargetMonitor`, respectively
* `DragSourceConnector` and `DropTargetConnector` are replaced by refs in hooks code
* Drag sources and drop targets don't take collect functions
* `useMonitorOutput` takes a monitor and a collect function, and returns what is returned by the collect function whenever the monitor changes
* `useDrayLayerOutput` is similar to `useMonitorOutput`, although only takes a collect function
* `connectDragTarget`/`connectDropTarget` is replaced with `ref`s
* `connectDragPreview` is replaced with `ref`s or dom nodes


packages/react-dnd/src/hooks/useMonitorOutput.ts
```diff
@@ -0,0 +1,63 @@
...
+	// This runs on every render. There will be ways to optimise this, but for
+	// now, this is the most correct thing to do.
+	React.useEffect(() => {
+		const handlerId = monitor.getHandlerId()
+		if (handlerId == null) {
+			return undefined
+		}
+		const options = { handlerIds: [handlerId] }
+		return (monitor as any).subscribeToStateChange(updateIfNeeded, options)
+	})
```
jamestalmage: Why not just add the empty array so it only runs once?

If we do need to run it every render, should we unsubscribe the previous instance?

packages/react-dnd/src/hooks/useMonitorOutput.ts
```diff
@@ -0,0 +1,63 @@
...
+	const [value, setValue] = React.useState(() => {
+		isFirstRender = true
+		return collect(monitor)
+	})
+
+	const updateIfNeeded = () => {
+		const nextValue = collect(monitor)
+		// Not async-safe, but we need a way to opt-out of state updates
+		if (!shallowEqual(value, nextValue)) {
+			setValue(nextValue)
```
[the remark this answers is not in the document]
jacobp100: We need a way to bail out of updates, which we don’t currently have

packages/react-dnd/src/hooks/useMonitorOutput.ts
jacobp100: I think this actually has to run every render

packages/react-dnd/src/hooks/useMonitorOutput.ts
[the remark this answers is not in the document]
jamiewinder: I think the stable release of hooks (16.8) had this?

packages/react-dnd/src/hooks/useDropTarget.ts
```diff
@@ -0,0 +1,85 @@
...
+		handler,
+		dragDropManager,
+		targetMonitor,
+	)
+
+	React.useEffect(() => {
+		const dropTargetNode = ref.current
+		const dropTargetOptions = targetSpec.dropTargetOptions
+
+		const disconnectDropTarget = backend.connectDropTarget(
```
jamiewinder: I think there needs to be a check here to ensure `dropTargetNode` is defined. For example, if the outer component attaches the ref to a conditionally rendered element then the backend (the HTML5 one at least) will attempt to add event listeners to 'null'.

packages/react-dnd/src/hooks/useDropTarget.ts
```diff
@@ -0,0 +1,85 @@
...
+	React.useEffect(() => {
+		targetSpecRef.current = targetSpec
+	})
+
+	// Can't use createSourceFactory, as semantics are different
+	const handler = React.useMemo(
+		() => ({
+			canDrop() {
+				const { canDrop } = targetSpecRef.current
+				return canDrop != null ? (canDrop as NoArgs<any>)() : true
```
jamiewinder: Since this doesn't pass the monitor to the callback, there is no way of knowing what is being dragged?

packages/react-dnd/src/hooks/useDropTarget.ts
jacobp100: Yeah, good point! I need to update this PR anyway to reflect the new Hooks API

packages/react-dnd/src/hooks/useMonitorOutput.ts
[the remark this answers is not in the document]
jacobp100: Yes, it does now!

packages/react-dnd/src/hooks/useDropTarget.ts
jacobp100: You might be right - I'm not sure at the moment. My intention was,

```js
const monitor = useDropTarget({
  canDrop() {
    monitor.getItem().type = 'something'
  }
})
```

darthtrevino:
The chessboard example looks great. I don't have write permission to the PR so I'm using my own branch. If you want to make changes to this, cut a PR against the pr/1173_mod branch

darthtrevino:
Closing in favor of #1244

Issue #1175, opened 2018-11-21
clientOffset is null in canDrag


0815Strohhut:
In the custom "canDrag" implementation, monitor.getClientOffset() (and monitor.getInitialClientOffset()) returns null.

In my application, I am using a DragSource around a large SVG where some parts should be draggable and others not - as it is used on touch devices as well, it is important that no drag operation is started for the points where drag is not possible to allow scrolling in those areas. As I don't have access to the internals of the SVG, the only possiblitly is to work with the coordinates and decide based on the coordinates which points are draggable and which aren't.

This follows up on https://github.com/react-dnd/react-dnd/issues/986. In the issue, the clientOffset/initialClientOffset and initialSourceClientOffset are null in the beginDrag. This has been fixed by implementing a new action "INIT_COORDS" that is called before the beginDrag and sets the coordinates to the dragState. However, as the initCoords already populates the "InitialSourceClientOffset", it uses the "canDrag" to determine which source is being dragged, thus the "canDrag" is called before the coordinates are set to the internal state. 

Based on my understanding of the current architecture, it wouldn't be easy to set the initialSourceClientOffset for every potential drag source. Thus, I am suggesting to add another action (INIT_CLIENT_OFFSET) that populates the client offset and the initial client offset while ignoring the sourceClientOffset - hence this method doesn't need to call "canDrag" and can populate the clientOffset before any "canDrag" is called. 

That means, the initialSourceClientOffset won't be available in the canDrag, but the clientOffset and initialClientOffset are available - allowing the respective drag source to calculate the sourceClientOffset itself should it be required.

This would also improve upon the following issues: 
https://github.com/react-dnd/react-dnd/issues/874
https://github.com/react-dnd/react-dnd/issues/519

I'll attach a PR in a sec.

PR #1176, opened 2018-11-21
fix: clientOffset and initialClientOffset are null in canDrag #1175

0815Strohhut:
See issue https://github.com/react-dnd/react-dnd/issues/1175

If I should add some tests/something is wrong please let me know.


darthtrevino:
I think this looks fine, but I wonder if it's worth using redux-thunk instead of using a separate action type

darthtrevino:
Ahh, the manager can dispatch to the store, so I think we can pull this off without adding a new action type. I'll have a PR up in a moment

Issue #980, opened 2018-02-28
Add license to all subrepos package.json


mlucool:
Hi,

Can you please update all package.json to include a license? It seems that the unpublished toplevel package.json has the correct one and all packages in the monorepo have a LICENSE file within them as well.

PR #1197, opened 2018-12-01
fix: #980 added MIT license to inner packages

sant0shg:
We have license files inside packages which are MIT license. But the license inside package.json was `BSD-3`. So changed the `BSD-3` to `MIT` where applicable to remove confusion, and added `MIT` license to packages without any license. @darthtrevino please let me know if the package.json had `BSD-3` for some particular reason.

### DIFF
--- a/packages/dnd-core/package.json
+++ b/packages/dnd-core/package.json
@@ -2,7 +2,7 @@
 	"name": "dnd-core",
 	"version": "7.0.1",
 	"description": "Drag and drop sans the GUI",
-	"license": "BSD-3-Clause",
+	"license": "MIT",
 	"main": "./lib/index.js",
 	"types": "./lib/index.d.ts",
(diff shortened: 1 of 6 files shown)


darthtrevino:
I think we should use BSD 3 clause in the inner packages.  MIT must have
been a mistake on my part. I prefer MIT, but in order to change the license
we would need Dan Abramov's signoff

gaearon:
The project has always been MIT: https://unpkg.com/react-dnd@0.1.0/LICENSE

I have no idea why BSD license was added. I don't think I did that? I see it first appearing in https://github.com/react-dnd/react-dnd/pull/892.

pd4d10:
Lerna project root(react-dnd-parent)'s license is still BSD. Shall we change it to MIT as well?

pd4d10:
#1198

Issue #998, opened 2018-04-03
Add support for React 16.3 createRef type of ref


janpe:
Using connectDragSource with an element that is using the new createRef API (introduced in React 16.3.0) instead of callback refs causes the app to crash with the error: `TypeError: previousRef is not a function`.

https://github.com/facebook/react/pull/12162

dantman:
Same issue. First time I finally get a chance to use the new refs and react-dnd is in the way.

Long term I'm not sure about the whole react-dnd model of using findDOMNode and replacing the ref. When React has new createRef and withRef. New [forwardRef API](https://reactjs.org/docs/forwarding-refs.html). And a number of APIs are starting to make use of render props.

But for now it should be sufficient to check if previousRef is an object with a `current` property and set that instead of calling it as a function.

MattBubernak:
Anyone have any solutions for this, while still using createRef()?

darthtrevino:
I can dig into this later, but a sandbox demonstrating what you're looking for would be great here.

dhruvparmar372:
I'm facing the same issue as well. 
@darthtrevino https://codesandbox.io/s/733onvqwl6 maybe this sandbox helps.

im-martijn:
Same issue. Should I revert to using the old callback ref, or is there a way to work around this?

markvital:
+1
using `React.createRef()` and `ref.current` crashed the app.

When this will be fixed?

AlexChrisF:
Still not fixed :/

darthtrevino:
I'll make some time after react-conf to look into this

arshmakker:
I was able to use a workaround as suggested by encapsulating the ref div inside another blank div.
example:
` <div>
        <div
          style={{
            position: "relative",
            borderStyle: "dashed",
            width: 300,
            height: 200,
            position: "relative"
          }}
          ref={this.myRef}
        />
      </div>`

Edmondton:
👍 any update on this @darthtrevino.  would be great to hear an update

piotrpawlik:
@darthtrevino Encountered same issue, could we get any ETA on the fix?

darthtrevino:
should be today I think. Let me know if there are any issues with the release I cut later

PR #1212, opened 2019-01-17
fix: work with elements that have Ref's generated using the Hooks API

jamestalmage:
With React Hooks, ref's no longer have to be functions, but can
be simple objects with a `current` property. Instead of calling the
`ref` function, React sets `ref.current`.

Currently, the following will throw an error:

```jsx
const MyComponent = ({connectDragSource}) => {
  const ref = useRef();

  return connectDragSource(<div ref={ref}/>);
};

DragSource(type, spec, connect)(MyComponent);
```

There is ongoing work to provide a Hooks API for react-dnd, which will ultimately be preferable, but this provides an immediate fix for mixing the current API with React Hooks.


RudyWeber:
Cool! I was about to submit a PR doing just that. Note that this change allows to use refs created using `React.createRef` (since React 16.3) which `useRef` uses and would close issue #998.

packages/react-dnd/src/utils/cloneWithRef.ts
```diff
@@ -24,7 +24,11 @@ export default function cloneWithRef(
 		ref: (node: any) => {
 			newRef(node)
 			if (previousRef) {
-				previousRef(node)
+				if (typeof previousRef === 'function') {
+					previousRef(node)
+				} else {
```
RudyWeber: Maybe
```javascript
else if (typeof previousRef === 'object' && 'current' in previousRef)
```
?
Just to be sure we are dealing with an actual ref.

jamestalmage: Sure.

airhorns:
This would be really great to get in if y'all had a minute or two, being able to use the new refs API would be very helpful!

darthtrevino:
So I merged a modified PR. As a clarification, `useRef()` will still not work out but `React.createRef()` should. Let me know if there are any issues with it.

PR #1220, opened 2019-02-05
fix #1146 ensure children mount first

### LINKED ISSUE #1081: v4.0.0 Breaks component lifecycles in app
**Describe the bug**
My app uses `componentDidMount` to measure an array of DOM elements. Those elements are wrapped in a `dragSource`
In v3.0.2, I could measure them in cDM. In v4.0.0, they all return 0s because they mount empty divs.
I dug through the diff and saw what I assume to be the offender: https://github.com/react-dnd/react-dnd/compare/v3.0.2...v4.0.0#diff-ac418ba19283aec1fb0b70e6570c5613R114

cDU is not a valid replacement for cWRP as it fires too late, although I'm not 100% sure that's the culprit. I'd like to pick your brain to see if you think something else may be the cause.
I know it isn't the HTML5 backend because i upgraded both & it was borked. I then removed the `DragSource` HOCs and it looked correct, so I'm relatively certain it's there in the `decorateHandler`

**To Reproduce**
i'll make a sample repo if needed, but should be pretty basic

**Expected behavior**
everything to work as it did in v3.0.2

**Screenshots**
N/A

 - OS: ubuntu
 - Browser chrome
 - Version 66

### LINKED ISSUE #1146: Children of drop target receive componentDidMount after parent
Normal React semantics is for child components to have `componentDidMount` called before parents. Since adding react-dnd I'm seeing components inside a `DropTarget` get this callback *after* parent components. I'm guessing there children are not rendered at all initially, and then added on a subsequent render?

Is this expected behaviour? Is there any way to be notified when the component is fully mounted, with all its children in place?

Thanks

mattkrick:
fixes #1146 fixes #1081

react guarantees that child components mount before their parents. react-dnd v4.0+ was breaking that contract. this fixes it.


mattkrick:
Hate to be a squeaky wheel, but any progress on this?

@cvlmtg i saw you had this problem too, if you could confirm that this works for you too, that may help move it along. Easy package.json drop-in:
```
"react-dnd": "https://github.com/mattkrick/react-dnd/tarball/2c22596327123a42702474374f6b0168c42f554b",
"react-dnd-html5-backend": "https://github.com/mattkrick/react-dnd/tarball/5812e3c00a1e1adace6e4be59bab7c02f0e85bb1",
```

cvlmtg:
@mattkrick it works, thanks!

PR #1223, opened 2019-02-11
Fix tutorial code snippets

mckomo:
I was going through the tutorial and noticed a few bugs in the code snippets. This PR aims to fix them.

**Changes are**:

* `Board` should not reference helper functions via `this` since it is an SFC.
* Declare and use `handleSquareClick`, `renderPiece` and `renderOverlay` as a static function to be consistent with the previous example where `renderSquare` was introduced as a static function.
* Use `DragDropContext` HOC instead of `DragDropContextProvider` to avoid `Cannot have two HTML5 backends at the same time.` error.


darthtrevino:
LGTM, I think we'll want to eventually make this more idiomatic (e.g. renderSquare is really a Function component, handleSquareClick should be a callback hook).

PR #1225, opened 2019-02-17
feat(html5-backend): expose `dataTransfer.items` for files

### LINKED ISSUE #712: HTML5 backend: access to DataTransfer.items when dropping files

renchap:
This commit allows to expose more than one property in the HTML5 backend. Now, `event.dataTransfer.items` are exposed when files are dropped, allowing the developer to handle dropping directories and getting their content.

I also documented the properties exposed for the built-in native types.

I tested it with a recent Chrome, Safari & Firefox and I have been able to drop directories and get their content. I also tested this version with my own app without issues.

Fixes #712

### DIFF
--- a/packages/documentation/markdown/docs/04 Backends/HTML5.md
+++ b/packages/documentation/markdown/docs/04 Backends/HTML5.md
@@ -1,11 +1,11 @@
 ---
-path: "/docs/backends/html5"
-title: "HTML5 Backend"
+path: '/docs/backends/html5'
+title: 'HTML5 Backend'
 ---
-*New to React DnD? [Read the overview](/docs/overview) before jumping into the docs.*
 
-HTML5
-===================
+_New to React DnD? [Read the overview](/docs/overview) before jumping into the docs._
+
+# HTML5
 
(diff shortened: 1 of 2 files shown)


darthtrevino:
@renchap I'm going to take a cut at improving the typings around this piece of code. I'm going to push to your PR branch in a sec

ClementeSerrano:
Hi @renchap ! I'm implementing DnD on a web app and when trying to access `dataTransfer.items` the API returns me nothing more than an object with a `length` prop equal to `0`. I'm trying to filter DnD uploads by folders or files but its not possible to access the data to apply that filtering.

I get the following from the browser (Chrome, last stable version):

```
DataTransferItemList {0: DataTransferItem, length: 1}
length: 0
__proto__: DataTransferItemList
```

Any idea to fix this?

Thanks in advance!!!!

epicmau5time:
@ClementeSerrano, did you ever get this resolved?

Issue #1245, opened 2019-03-07
Crashing in isOver with empty targetId


Kadrian:
**Describe the bug**
I have a connected drop target that is removed from the DOM, once something is dropped on to it.

Now since upgrading to `v7.1.0`, I'm getting
```
Uncaught TypeError: Cannot read property '0' of undefined
```
in the `monitor.isOver()` call inside the `collect` function. 
Somehow, it sees that the collect function is still called one more time, after the drop target has been removed (`targetId` is undefined).

**Interesting**
- Downgrading back to `6.0.0` resolved the issue.
- Also, using a short `setTimeout` wrapper around my callback that removed the drop target, resolved the issue.

Any idea what could cause that?

**To Reproduce**
1. Connect a Drop target, call `monitor.isOver()` in `collect`
2. in the target, dispatch an action inside `drop(...)` that removes the drop target
3. See error

**Expected behavior**
There is no timing issue with `isOver` and drop targets can be immediately removed.

**Screenshots**
<img width="458" alt="bildschirmfoto 2019-03-07 um 00 58 04" src="https://user-images.githubusercontent.com/1403093/53922542-19d12900-4075-11e9-963b-0e9a1279dcb2.png">

**Desktop (please complete the following information):**
 - OS: os x
 - Browser chrome
 - react-dnd / html5 backend `7.1.0`

RaoHai:
Same issue

RaoHai:
I tried to revert this change https://github.com/react-dnd/react-dnd/commit/0feb250b7ee90483e31f3bc159ebf946980d53a7#diff-ac418ba19283aec1fb0b70e6570c5613 and resolved.. .

safonoi:
Same issue.
Waiting for fix.
Now downgraded back to v7.0.2

cfrank:
Confirmed we are receiving the same issue - Reverting to 7.0.2 fixes the issue

cfrank:
This is being caused because the change at 0feb250b7ee90483e31f3bc159ebf946980d53a7 is allowing target ID to be `undefined` but nobody checked the PR with that change for `isOverTarget`

Currently it's prototype is:

`public isOverTarget(targetId: string, options = { shallow: false })`

unlike the other functions which changed in that pr which move `targetId` / `sourceId` to

`targetId: string | undefined`
`sourceId: string | undefined`

and check:

```
if (!targetId) {
    return false;
}
```

[...]

PR #1248, opened 2019-03-07
targetId undefined on initial render being passed to parseRoleFromHandlerId 

cfrank:
fixes: #1245 

In a previous change set targetId/sourceId was changed so that it could
be passed undefined to internal functions. Not all functions were
changed to check for this change causing parseRoleFromHandlerId to
reference an undefined variable.

Signed-off-by: Chris Frank <cfrank@redventures.com>


darthtrevino:
Thanks for the PR! I modified it slightly on my end, but I'll release it soon

PR #1255, opened 2019-03-08
Hooks API Updates

darthtrevino:
__Naming__
`useDragSource` => `useDrag`
`useDropTarget` => `useDrop`

__Specifications__
* Move ref object and type parameters into spec object on both
`useDrag({ref, type, ...})`
`useDrop({ref, type, ...})`

Moving some examples over to hooks as a test-bed. Currently the simple-sortable example can get out of sync with the dragdrop state 

Reverting the documentation to using the class-based API so that newcomers do not think the hooks API is idiomatic at this point.

NOTE:
The hooks API will be placed inside of a super annoying namespace in react-dnd until it's ready for production:
```
import {__EXPERIMENTAL_DND_HOOKS_} from 'react-dnd'
const { useDrag, useDrop} = __EXPERIMENTAL_DND_HOOKS_
```


brason:
Great so see support for hooks! Do you have an idea of how long till it's production ready?

darthtrevino:
Thanks! I don't have a fixed time-range for when the hooks-based API will be ready, but the main things I'd like to see are:

* Some more feedback on the hook-based API (it's still very fluid at this point)
* The hook-versions of the examples all have parity to the decorator-based versions. Right now the sortable-list example acts a little wonky and the "stress test" sortable list is noticeably slower than the decorator based version.

PR #1267, opened 2019-03-13
New Feature Proposal: Connect with Ref Objects

darthtrevino:
The current mechanism for wiring react elements into the backends seems dated. It was designed before React ref objects were conceived of in React.

The proposed mechanism here will be more straightforward, and will mirror what we see in the hooks-based API.  

The `dragSourceConnector` has the following new properties: `dragSourceRef`, `dragPreviewRef`.
The `dropTargetConnector` has the following new property: `dropTargetRef`

If the existing `connect<X>()` functions are used, they will take precedence over the ref objects.

__Basic Example__
```js
function Card({ dragSource, text }) {
  return(
    <div ref={dragSource}>
      {text}
    </div>
  );
}
export default DragSource(
  ...
  (connect, monitor) => ({
    dragSource: connect.dragSourceRef,
  })
)(Card);
```

__Composite Example with Dnd-Provided Ref__
```js
function Card({ dragSource, connectDropTarget, text }) {
  connectDropTarget(dragSource);
  return(
    <div ref={dragSource}>
      {text}
    </div>
  );
}

export default DragSource(
    ...
  (connect, monitor) => ({
      dragSource: connect.dragSourceRef,
    })
)(DropTarget(
 ...,
 (connect) => ({
    connectDropTarget: connect.dropTarget()
  })
)(Card);
```

__Composite Example with Client-Provided Ref__
```js
function Card({ connectDragSource, connectDropTarget, text }) {
  const ref = useRef(null);
  connectDragSource(ref)
  connectDropTarget(ref);
  return(
    <div ref={ref}>
      {text}
    </div>
  );
}

export default DragSource(
    ...
  (connect, monitor) => ({
      dragSource: connect.dragSource()
    })
)(DropTarget(
 ...,
 (connect) => ({
    connectDropTarget: connect.dropTarget()
  })
)(Card);
```

### DIFF
--- a/packages/dnd-core/package.json
+++ b/packages/dnd-core/package.json
@@ -10,8 +10,7 @@
 		"build:cjs": "tsc -b tsconfig.cjs.json",
 		"build": "run-p build:*",
 		"clean": "rimraf lib",
-		"watch": "tsc -w --preserveWatchOutput",
-		"start": "npm run watch",
+		"start": "tsc -b tsconfig.cjs.json -w --preserveWatchOutput",
 		"test": "run-s clean build"
 	},
 	"repository": {
--- a/packages/documentation/package.json
+++ b/packages/documentation/package.json
@@ -49,8 +49,7 @@
 	},
 	"scripts": {
 		"clean": "rimraf .cache public apidocs static/examples_js",
-		"develop": "gatsby develop",
-		"start": "npm run develop",
+		"start": "gatsby develop",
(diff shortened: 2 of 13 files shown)


darthtrevino:
Using refs instead of connect() functions makes sense, but we may need to think through how this affects certain use cases: e.g.: composite usage of the functions. A common pattern is to use connectDragSource and connectDropTarget together.

We'll probably need to allow injection of refs similar to the hooks API.

PR #1270, opened 2019-03-15
Don't create native drags out of an array, and make them enumerable

gustavohenke:
The `Object.create()` thing was a weird one. It really meant to be a plain object, I think.

Then I made the properties enumerable so we can do stuff like `{ ...monitor.getItem() }` more easily 😄


darthtrevino:
I'll try to make some time this week to test it out. It looks reasonable at first glance though

darthtrevino:
LGTM!

PR #1280, opened 2019-03-20
Hooks Ref'ing Updates

darthtrevino:
## Key Updates to the Hooks API per our recent discussion in #1148 

* Remove `ref` key on dragSource, dropTarget spec.
* `useDrag` and `useDrop` return connector functions, which can now be chained and use as refs
* ~~Change the top-level useDrag and useDrop functions to accept a function and memoize keys so that the spec objects aren't created on every render.~~

__Simple Case__
```js
let [, drag] = useDrag({/*...*/})
return <div ref={drag}>...</div>
```

__Chained Case__
```js
let [, drag] = useDrag({/*...*/})
let [, drop] = useDrop({/*...*/})
return <div ref={node => drag(drop(node))}>...</div>
```
or
```js
let ref = useRef(null)
let [, drag] = useDrag({/*...*/})
let [, drop] = useDrop({/*...*/})
drag(drop(ref))
return <div ref={ref}>...</div>
```

## Drag Previews
Remove `preview` key on dragSource spec, use connector function instead
~~Remove return ref from `useDragPreview`~~
~~Rename `useDragPreview` to `useDetachedComponent`~~
Add a DragPreviewImage component to render drag preview images on the output tree. (Note: The useDragPreview mechanism was removed inherit limitations in the HTML5 drag-and-drop API (https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/setDragImage) limit us to images)

```js
let [, drag, preview] = useDrag(() => ({/*...*/}))
return (
    <div ref={drag}>
        <DragPreviewImage connect={preview} src="imageSrc.png" /> 
        ...
    </div>
)
```
## Known Issues
- [ ] the shallowEquals check in `useCollector` is necessary for performance, but right now the hooks don't render correctly at all with it, and I'm not sure why.
- [ ] the draglayer example is wonky, dragged item is not disappearing and custom drag layer is wrong size
- [ ] the sortable card example is wonky, dragged item gets out of sync with view
- [ ] the stress test example does not function due to issue above.


packages/documentation/markdown/docs/05 Hooks-Based API/useDetachedComponent.md
```diff
@@ -16,16 +16,16 @@ import { __EXPERIMENTAL_DND_HOOKS_THAT_MAY_CHANGE_AND_BREAK_MY_BUILD__ as dnd }
 const { useDragPreview } = dnd
 
 function DragLayerPreview(props) {
-  const [DragPreview, preview] = useDragPreview(spec)
-  const [collectedProps, ref] = useDrag({
+  const DragPreview = useDetachedComponent(spec)
```
k15a: Any reason for this name change?

darthtrevino: It's not technically restricted to drag previews, it's just a utility to automate the boilerplate of rendering a ref-forwarding component into a detached portal.

packages/documentation/markdown/docs/05 Hooks-Based API/useDetachedComponent.md
```diff
@@ -16,16 +16,16 @@ import { __EXPERIMENTAL_DND_HOOKS_THAT_MAY_CHANGE_AND_BREAK_MY_BUILD__ as dnd }
...
+  const [, drag, preview] = useDrag({
     item: { id, type },
     preview,
```
k15a: You are using the preview which you return as an argument. This doesn't seem to make a lot of sense to me.

Also is there any special reason why the item is now `{ id, type }` and the `useDrag` accepts only one parameter instead of two  `useDrag(item, spec)`?

packages/documentation/markdown/docs/05 Hooks-Based API/useDetachedComponent.md
k15a: Anyting which speaks against `<DetachedComponent {...spec} />`? Sorry if you discussed that already somewhere.

packages/documentation/markdown/docs/05 Hooks-Based API/useDetachedComponent.md
darthtrevino: The DragPreview returned from `useDetachedComponent` is a component to render as a drag preview.

The `preview` returned from `useDrag` is a connector function that wires a node into react-dnd.

The item field was change to its current form for clarity. My goal with hooks has been to simplify the dnd API as much as is feasible.

k15a: What I meant is that the code in the example will throw an error as you are using the preview variable before it's defined:

> can't access lexical declaration `preview' before initialization

```
  const [, drag, preview] = useDrag({
    item: { id, type },
    preview,
  })
```

packages/documentation/markdown/docs/05 Hooks-Based API/useDetachedComponent.md
darthtrevino: Eh, I'll it looks it'll be preview-specific. I'll probably name it `useDetachedPreview`

packages/documentation/markdown/docs/05 Hooks-Based API/useDetachedComponent.md
darthtrevino: Oh, right. Documentation issue. `preview` isn't on the drag spec anymore.

darthtrevino:
I'm not really a fan of using a callback to memoize. I'd like to run some tests to see if a function arg really makes a difference here.

packages/documentation/markdown/docs/05 Hooks-Based API/useDetachedComponent.md
darthtrevino: Well, you can use the preview connector on any node. So you could do something like this: 

```js
<div ref={preview}>
    <div ref={drag}>//Drag Handle//</div>
</div>
```

The custom hook for disconnected previews is just to make those situations feel more at home in React. All the examples previously used Image tags directly via DOM apis. You can still do that, or you could do your own rendering into a portal and use that.

k15a: Still, don't understand what benefit of having a hook instead of a component. I don't think that hooks are the right choice here. But maybe that's just personal preference and a lot of bikeshedding.

```
// Higher order component

const DragPreview = createDetachedComponent(KnightDragPreview);

function Component() {
  return <DragPreview ref={ref} />;
}

// Hook

function Component() {
  const DragPreview = useDetachedComponent(KnightDragPreview);
  return <DragPreview ref={ref} />;
}

// Simple component

function Component() {
  return <DragPreview ref={KnightDragPreview} component={DragPreview} />;
}
```

As we are basically creating a new component based on another component the HOC version would make the most sense to me. If we don't create a new component and we just render two different ones the Simple component version would make the most sense to me.

darthtrevino: That's fair, if it's a Hoc then hoc-based API users can utilize this as well

darthtrevino: That's a fair point.  I'm not sure whether either approach has more technical merit, but if this is a HOC, then users of the HOC API can benefit from it, and not just users who are using hooks.

darthtrevino: After testing this a bit and trying to use arbitrary DOM structures, I realized that you can only connect images to the HTML5 dnd API, so I've removed `useDetachedComponent` and the HOC version and have replaced them with a plain component.

packages/react-dnd/src/DragDropContext.tsx
```diff
@@ -103,10 +102,8 @@ export function DragDropContext(
 			public render() {
 				return (
 					<Provider value={childContext}>
-						<Decorated
-							{...this.props}
-							ref={isClassComponent(Decorated) ? this.ref : undefined}
```
yched: The removal of the "isClassComponent(Decorated)" logic causes a "Function components cannot be given refs" error 
in 7.4.0 when DragDropContext is used to wrap a non-class component, which was supported so far. (I'm guessing this also affects other parts of the API where the isClassComponent() check was removed)

darthtrevino: I realized that and added the check back in

darthtrevino: Ahh, I see, I only restored the check in the decorateHandler. I'll restore those checks in a patch version today.

yched: Awesome, thanks !

PR #1283, opened 2019-03-20
test: added unit tests for react-dnd-html5-backend

sant0shg:
Added unit tests for react-dnd-html5-backend files.
There are some private methods in class/functions inside functions which cannot be directly tested. I have not made any code refactor in source files to make it easily testable. This PR is to add the test cases for code which can be easily tested.


darthtrevino:
Thank you @sant0shg !

PR #1284, opened 2019-03-21
New Feature: Add props to collector function arguments

### LINKED ISSUE #1282: Accessing target props within collect function
Is there any way to access target item props within `collect` function of `DropTarget`? The problem is that I need to show some specific message when user tries to drop the source into target that doesn't match `canDrop` criteria.
I have tried to do it by moving validation from `canDrop` to `dragEnd` but I still need to dynamically determine `isHighlighted` property using `canDrop` to highlight available containers for user. I have tried to use the following approach but it allows me to show just some generic message to user in case validation failed (because `monitor.getDropResult()` returns `null` in case when `canDrop` returns `false` and therefore `monitor.didDrop()` returns `false` as well.

Any ideas how to solve this issue? Thanks.

```javascript
export const Container = DropTarget(
  'container',
  {
    drop({ container }: Props) {
      return { container };
    },
    canDrop(props: Props, monitor) {
      const { box } = monitor.getItem();

      return !props.container.box && props.container.color === box.color;
    },
  },
  (connect, monitor) => ({
    isHighlighted: monitor.canDrop(),
    isOver: monitor.isOver(),
    connectDropTarget: connect.dropTarget(),
  }),
)(ContainerComponent);
```

```javascript
const Box = DragSource(
  'container',
  {
    beginDrag({ box }: Props) {
      return { box };
    },

    endDrag({ t, attachAssetToTransaction, logId }: AssetProps, monitor) {
      if (!monitor.didDrop()) {
        // Using somewhat reliable internal method to determine if the box was actually dropped on any container
        if ((monitor as any).getTargetIds().length) {
          toast.error('Some generic error message here');
        }

        return;
      }
      const { box } = monitor.getItem();
      const { container } = monitor.getDropResult();

      putBoxIntoContainer(box.id, container.id);
    },
  },
  (connect, monitor) => ({
    connectDragSource: connect.dragSource(),
    isDragging: monitor.isDragging(),
  }),
)(({ connectDragSource, isDragging, ...rest }: BoxProps & DragProps) =>
  connectDragSource(
    <div
      className={classNames(s.awaitingAsset, { 'is-dragging': isDragging })}>
      <Box isDragging={isDragging} {...rest} />
    </div>,
  ),
);
```

darthtrevino:
fixes #1282 

To test this feature , I added a new package: `react-dnd-test-utils` to house the `wrapInTestContext` function.


darthtrevino:
GAH!

darthtrevino:
Too early!

PR #1293, opened 2019-03-26
Remove dependency on recompose

### LINKED ISSUE #1229: Yarn install error: "Couldn't find any versions for "recompose" that matches "^0.30.0""
**Describe the bug**
When I try to install `react-dnd` via `yarn`, I get the following error: `Couldn't find any versions for "recompose" that matches "^0.30.0". Possible versions: "0.20.2, 0.23.5, 0.26.0"`

I looked on the `yarn` website to see if maybe `0.30.0` isn't included in the registry, but indeed it is, and it's been around for six months.

**To Reproduce**
1. Add `react-dnd` version `7.0.2` to `dependencies` in `package.json`.
2. `yarn install -force`
3. Observe the error.

**Expected behavior**
1. Add `react-dnd` version `7.0.2` to `dependencies` in `package.json`.
2. `yarn install -force`
3. Victory is ours

 - OS: Windows 10
 - Browser: n/a
 - Version: `7.0.2`

inukshuk:
This removes the "recompose" dependency [by copying `isClassComponent` from the recompose module](https://github.com/acdlite/recompose/blob/master/src/packages/recompose/isClassComponent.js); the predicate itself is trivial and recompose is not actively maintained anymore so this shaves off over 8 MB, or more than half of all production dependencies when installing via NPM. (I know most people will use bundlers anyway, but the extra weight still hurts in some situations.)

As far as I can tell `isClassComponent` was the only part of recompose that was actually used; I hope I did not miss anything!

Before:
Screenshot from 2019-03-26 12-50-04

After:
Screenshot from 2019-03-26 12-55-17

Although unrelated, should close #1229 

Thanks for react-dnd!

### DIFF
--- a/packages/react-dnd/package.json
+++ b/packages/react-dnd/package.json
@@ -25,7 +25,6 @@
 		"hoist-non-react-statics": "^3.3.0",
 		"invariant": "^2.1.0",
 		"lodash": "^4.17.11",
-		"recompose": "^0.30.0",
 		"shallowequal": "^1.1.0"
 	},
 	"devDependencies": {
--- a/packages/react-dnd/src/decorateHandler.tsx
+++ b/packages/react-dnd/src/decorateHandler.tsx
@@ -11,11 +11,17 @@ import {
 	SerialDisposable,
 } from './utils/disposables'
 import { Connector } from './SourceConnector'
-const isClassComponent = require('recompose/isClassComponent').default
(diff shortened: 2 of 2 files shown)


darthtrevino:
You're right though, there's no great reason for the dependency. Since recompose was moved to the `isRefable` utility on master, I went ahead and did this manually

inukshuk:
Awesome, thanks!

PR #1330, opened 2019-04-20
fix(typescript): declare hoist-non-react-statics type dependency

jleclanche:
react-dnd directly uses hoist-non-react-statics in its type declarations, which makes it a direct dependency and unfortunately, according to https://github.com/mridgway/hoist-non-react-statics/issues/47, that repo is refusing to package type definitions.

This PR fixes the issue for typescript users that don't happen to have their own dependency on `@types/hoist-non-react-statics`.

image


spassvogel:
Would be great to have this approved!

PR #1355, opened 2019-05-27
fix: always unsubscribe backend when connector reconnects

### LINKED ISSUE #1349: Memory leak for unmounted components
**Describe the bug**
When the DragSource-component is unmounted, the unsubscribe function (https://github.com/react-dnd/react-dnd/blob/master/packages/react-dnd-html5-backend/src/HTML5Backend.ts#L121) is not called.

**To Reproduce**
1. Open example from link: https://codesandbox.io/s/memory-leak-reactdndhtml5backend-64y8r
or manually execute this code:
```js
import React from "react";
import ReactDOM from "react-dom";
import { DragDropContextProvider, DragSource } from "react-dnd";
import HTML5Backend from "react-dnd-html5-backend";

const Draggable = DragSource(
  "",
  {
    beginDrag() {},
    endDrag() {}
  },
  connect => ({
    connectDragSource: connect.dragSource()
  })
)(({ connectDragSource }) => <div ref={connectDragSource} />);

let backend;

class App extends React.Component {
  constructor(props) {
    super(props);
    this.state = { cnt: 1 };
    setInterval(() => this.setState({ cnt: this.state.cnt + 1 }), 10);
  }

  render() {
    return (
      <>
        sourceNodes in memory: {backend.sourceNodes.size}
        {this.state.cnt % 2 === 0 && <Draggable />}
      </>
    );
  }
}

ReactDOM.render(
  <DragDropContextProvider backend={m => (backend = HTML5Backend(m))}>
    <App />
  </DragDropContextProvider>,
  document.getElementById("root")
);
```
2. The number of sourceNodes in memory strictly increases to infinity.

**Expected behavior**
The number of sourceNodes does not increase above the actual number of components.

**Screenshots**
test

**Desktop:**
 - OS: Windows 10
 - Browser: Chrome
 - Version: 74

**Additional context**
 - react: 16.8.6
 - react-dom: 16.8.6 
 - react-dnd: 7.4.5
 - react-dnd-html5-backend: 7.4.4

upsuper:
This fixes #1349.

### DIFF
--- a/packages/react-dnd/src/SourceConnector.ts
+++ b/packages/react-dnd/src/SourceConnector.ts
@@ -93,18 +93,22 @@ export default class SourceConnector implements Connector {
 	}
 
 	private reconnectDragSource() {
+		// if nothing has changed then don't resubscribe
+		const didChange =
+			this.didHandlerIdChange() ||
+			this.didConnectedDragSourceChange() ||
+			this.didDragSourceOptionsChange()
+
+		if (didChange) {
+			this.disconnectDragSource()
+		}
+
 		const dragSource = this.dragSource
 		if (!this.handlerId || !dragSource) {
 			return
 		}
 
-		// if nothing has changed then don't resubscribe
-		if (
-			this.didHandlerIdChange() ||
-			this.didConnectedDragSourceChange() ||
-			this.didDragSourceOptionsChange()
-		) {
-			this.disconnectDragSource()
+		if (didChange) {
 			this.lastConnectedHandlerId = this.handlerId
 			this.lastConnectedDragSource = dragSource
 			this.lastConnectedDragSourceOptions = this.dragSourceOptions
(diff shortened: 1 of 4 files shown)


BrendanAnnable:
@darthtrevino any chance of getting this fix expedited into a release ASAP?

This is currently causing major memory leaks and crashes on our production servers due to the disposers not being called on component unmount. We'd really appreciate a patch release to avoid forking 🙂

darthtrevino:
Sure, I'll have a look tonight

BrendanAnnable:
Thanks for merging! Could we get an npm release out as well? 🙏 😅

darthtrevino:
Yep, I'll go through releasing it tonight

darthtrevino:
Hey Brendan,
I was going through the examples tonight and unfortunately I can't release this quite yet.

If you start the app, and then open http://localhost:8000/examples/customize/drop-effects, you'll see that example does not work. In fact, any example that uses the `connectDragSource(<dom-node>` pattern will not work here. I'll need to dig into this further to understand what's going on. 

As an aside, this may be an excuse to deprecate that pattern and cut a major semver that forces clients to pass ref objects to the connector functions.

darthtrevino:
https://github.com/react-dnd/react-dnd/pull/1374 this fixes the issue in that particular example, but this example is a problem: http://localhost:8000/examples/dustbin/stress-test

It looks like the elements are draggable out-of-the gate, but lose that ability when the UI refreshes

upsuper:
Hey @darthtrevino, any news on the new release?

paulshen:
i opened #1379 which seems to be needed as well

PR #1366, opened 2019-05-31
fix: declare module on dnd-core

### LINKED ISSUE #1365: react-dnd is not tree-shakable
**Describe the bug**
The distributed npm package includes both `cjs` and `esm` builds, but `package.json` only declares `main` property and not `module` property.
Therefore, when building with a tree-shaking supported bundler (webpack, for instance), react-dnd is packaged completely and without tree-shaking.

**To Reproduce**
1. Create an app with `react-dnd`
2. Build using webpack production mode
3. Analyze output, all assets are there

**Expected behavior**
`react-dnd` is tree-shaked at build time and only imported assets should be on final bundle.

aldo-roman:
fixes #1365

### DIFF
--- a/packages/dnd-core/package.json
+++ b/packages/dnd-core/package.json
@@ -4,6 +4,7 @@
 	"description": "Drag and drop sans the GUI",
 	"license": "MIT",
 	"main": "./lib/cjs/index.js",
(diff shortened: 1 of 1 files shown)


aldo-roman:
Seems to have been ommitted from #1319

PR #1415, opened 2019-06-20
Fix tutorial code

hansnow:
Add the missing drop ref, so that `BoardSquare` could be a valid DropTarget.

### DIFF
--- a/packages/documentation/docsite/markdown/docs/00 Quick Start/Tutorial.md
+++ b/packages/documentation/docsite/markdown/docs/00 Quick Start/Tutorial.md
@@ -596,6 +596,7 @@ function BoardSquare({ x, y, children }) {
 
   return (
     <div
+      ref={drop}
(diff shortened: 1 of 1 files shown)


darthtrevino:
If you can undo the part where you replace DndProvider with DragDropContext, I can merge this. I want to deprecate that part of the API.

hansnow:
Of course, I can undo that part. But for now, `DndProvider` will cause `Cannot have two HTML5 backends at the same time.` error in tutorial code snippets. Replacing `DndProvider` with `DragDropContext` is the minimum fix to make the tutorial code work again. If you insist on using `DndProvider`, the `observe` pattern should be rewritten.

darthtrevino:
The two-backend issue has been resolved in 8.0.3

hansnow:
Roger that. Have reset the related commit.

PR #1429, opened 2019-06-26
fix: stale isDropping on drag source unmount/remount (hooks)

qtsd:
Here is a [codesandbox](https://codesandbox.io/s/stale-isdragging-on-drag-source-unmountremount-using-hooks-nj3ip) to see the problem.

The drag source is registered on an effect, so the first render doesn't have the correct state.
And since the element is already dragging, the store subscription is not triggered.

### DIFF
--- a/packages/core/react-dnd/src/hooks/internal/drag.ts
+++ b/packages/core/react-dnd/src/hooks/internal/drag.ts
@@ -71,7 +71,7 @@ export function useDragHandler<
 		} as DragSource
 	}, [])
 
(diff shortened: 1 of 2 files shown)


qtsd:
In fact, useMemo is not the solution, since it could be recalled anytime.
I think it's better to manually recollect after the registration.

darthtrevino:
Thanks for the PR @qtsd - I wanted to add the codesandbox as a doc example, so I'm closing this out in favor of #1437

PR #1491, opened 2019-07-26
fix path to cjs bundle instead esm bundle

ramilzakirov:


HsuTing:
Does this `pr` have any problem? Will this `pr` is planned to be merged?

ramilzakirov:
> Does this `pr` have any problem? Will this `pr` is planned to be merged?

I would like this PR to be accepted! This fixes a unit testing error with jest.

HsuTing:
@raineorshine `react-dnd-html5-backend` also has the same problem, can you fix this in this `pr`?
If you can double check `package.json` in this repo, will be better.

darthtrevino:
This change was intentional in v8 when we switched to ESM. 

If you need to make Jest work, check out the testing docs: http://react-dnd.github.io/react-dnd/docs/testing

If you need CJS support for reasons outside of Jest, consider using either a Webpack alias or using the `react-dnd-cjs` package.

HsuTing:
@darthtrevino I just want to use `react-dnd` in the server side, but `nodejs` is not supported to use `ESM`, now.

I have no idea why both of `pkg.main` and `pkg.module` should use `esm`. If the tool is supported to use `ESM`, like `Webpack`, it will read the `module` in the `package.json` first. As a result, I think we should use `cjs` in `pkg.main` and `esm` in `pkg.module`?

darthtrevino:
pkg.module doesn't appear to be part of the NodeJS standard there. (https://nodejs.org/api/esm.html#esm_package_entry_points). Using CJS as the entrypoint would be a major semver. change.

Issue #1532, opened 2019-09-11
Memory leak with dragDropManager as retainer object


canastro:
**Describe the bug**

Possible memory leak with react-dnd. 

In a app with react-router-dom when I navigate away of a screen with react-dnd I still see a lot of my models instances, unrelated components (not child neither sibilings of my component containing the DnDProvider) and `detached dom elements` with references to `dragDropManager`. 

At the current state the memory usage after I leave the screen with the DnD is around 49MB, if I comment out the DND components (which is now as simple as possible: provider + 1 drop zone) it goes to 28MB.

I was able to recreate a small sample which only shows the leak of detached dom elements with the references to `dragDropManager`.

I couldn't yet replicate the scenario where multiple components are kept in memory with dragDropManager as retainer object. But here is a picture of the retainer in one of my components:
<img width="624" alt="Screenshot 2019-09-11 at 10 37 34" src="https://user-images.githubusercontent.com/1267623/64686279-55173f00-d480-11e9-802a-80d725ab6898.png">

**Reproduction**

**Video showing the possible memory leak:** https://monosnap.com/file/1aKyq523ZkKY1gO39QS05yjBmzEBB0

1. Clone https://github.com/canastro/dnd-sample
3. Run sample in production mode with `npm run prod`
4. Open http://localhost:5000
5. Go to the DND page and navigate back and sometimes between the other two pages
6. Take heap snapshot, check for dettached div elements, and check a retainer related with react-dnd

**Expected behavior**
All references to react-dnd should be cleaned up properly.

**Desktop (please complete the following information):**

- OS: macOS Movaje
- Browser: chrome
- Version:  76.0.3809.132

canastro:
@gaearon since you're familiar with both react an react-dnd can you point out if this issue is something related with DND or if it fits the points already identified in the [memory leaks umbrella issue of react](https://github.com/facebook/react/issues/16087)?

canastro:
I've found two different ways to address this, one in react-dnd's source code (but probably this would mess up 1001 different use cases) and another in user-land.

In react-dnd source code I would add something roughly like this in the DndProvider:

```js
React.useEffect(() => {
	return () => {
		if ('manager' in props) {
			return;
		}

		const context = props.context || getGlobalContext();
		context[instanceSymbol] = null;
	}
}, []);
```

In user land I would have to use the DND in one of two following ways:

```jsx
const manager = useRef(createDndContext(HTML5Backend));

[...]

PR #1543, opened 2019-09-20
Keep track of number of references to global context and cleanup

canastro:
Tries to address https://github.com/react-dnd/react-dnd/issues/1532


packages/core/react-dnd/src/common/DndProvider.tsx
```diff
@@ -12,26 +12,59 @@ export type DndProviderProps<BackendContext, BackendOptions> =
 			context?: BackendContext
 			options?: BackendOptions
 			debugMode?: boolean
-	  }
+		}
+
+let globalContextRefCount = 0;
```
canastro: is there a way I can use the `refCount` from react-dnd's store? I'm not sure if this globalContextRefCount is the best approach.

package.json
```diff
@@ -46,7 +46,7 @@
 		"release:packages": "lerna publish",
 		"go_no_go": "node scripts/go_no_go.js",
 		"release_docs": "lerna run release --stream --scope react-dnd-documentation",
-		"test": "run-s clean lint build test_modules jest:cov",
```
darthtrevino: Revert this line

Issue #1506, opened 2019-08-07
Enzyme testing instruction out of date


sandorfr:
**Describe the bug**
Writing test with enzyme as stated here will not work http://react-dnd.github.io/react-dnd/docs/testing#testing-with-enzyme

**Reproduction**

Write a test using enzyme, 
Try to retrieve the manager
```javascript
var backend = root
  .instance()
  .getManager()
  .getBackend()
```
It will fail because instance() will return null with React 16 (Stateless function component don't have an instance)

**Expected behavior**
The doc should provide the recommanded working way to achieve this.

**Workaround**
I figured something like this, but does feel quite convoluted:
```
    const mountComponent = (Component: any, testProps: TestProps) => {
        const result: any = { wrapper: undefined, manager: undefined };

        result.manager = new Promise<DragDropManager>(resolve => {
          
            result.wrapper = mount(
                <DndProvider backend={TestBackend}>
                    <DndContext.Consumer>
                        {dnd => {
                            resolve(dnd.dragDropManager);
                        }}
                    </DndContext.Consumer>
                    <Component
                        {...testProps}
                    />
                </DndProvider>,
            );
        });
        return (result: {|
            wrapper: ReactWrapper<DndProvider>,
            manager: Promise<DragDropManager>,
        |});
    };
```
- OS: MacOS
- Browser: Chrome
- Version : 75
- React: 16.8.6
- React-dnd : 9.3.9

RuGrof:
How can I get getHandlerId method for my component?

sa-scole:
`.getHandlerId()` is an instance method for getting the sourceId for DropTarget, which is part of the Legacy Decorator API. It may be that this part hasn't been updated

sa-scole:
@darthtrevino Do you have any guidance here?

JKillian:
Hit this myself today. The library can internally just use `forwardRef` / `useImperativeHandle` to expose the methods in functional components, right?

JKillian:
Found a workaround that might be slightly cleaner than @sandorfr's above:

First, define this helper for yourself:

```ts
  const wrapInTestContext = Component =>
    React.forwardRef((props, ref) => {
      const dragDropManager = useRef();
      useImperativeHandle(ref, () => ({ getManager: () => dragDropManager.current }));

      return (
        <DndProvider backend={TestBackend}>
          <DndContext.Consumer>
            {ctx => {
              dragDropManager.current = ctx.dragDropManager;
            }}
[...]

PR #1570, opened 2019-10-07
fix: update wrapInTestContext to expose `getManager()` method

JKillian:
Tweaks the `wrapInTestContext` [as per this comment](https://github.com/react-dnd/react-dnd/issues/1506#issuecomment-537699644) so that it gives consumers an easy way to access the `DragDropManager` in their tests.

Also updates the testing instructions for Enzyme to show how to work around an enzyme bug in order to access the `DragDropManager`.


JamesMHenderson:
How would you simulate the drag events?

JKillian:
> How would you simulate the drag events?

Further changes to other components might also be necessary to make that possible, I haven't looked in to it. But figured might as well get the ball rolling on making things more testable with Enzyme!

NYannick:
Hello any update on this fix ?

darthtrevino:
LGTM!

PR #1577, opened 2019-10-12
Declared packages side effects free

maclockard:
This sets the `package.json` field `sideEffects` to `false`. This allows for better tree shaking by downstream consumers, see https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free for more information.

I skimmed over all included files and I did not see anything that would count following webpack's definition of tree shaking during a production build.

I did however find a couple side effects, but they appear to only be relevant during development, which treeshaking is not run during. Here are the notable examples:
- https://github.com/react-dnd/react-dnd/blob/master/packages/core/react-dnd/src/common/DragPreviewImage.tsx#L21
  - There are a couple examples of this, my understanding is this is just to set the name for react dev tools, so not relevant in production.
  - This is fixable if you would like by using `React.memo` as the documentation suggests with a named function: https://reactjs.org/docs/react-api.html#reactmemo
 - https://github.com/react-dnd/react-dnd/blob/master/packages/core/dnd-core/src/utils/dirtiness.ts#L6-L7
    - My understanding here is this is just used for debugging and is okay to drop


darthtrevino:
I really like the idea of deeper tree-shaking support, but I'm reticent to declare all the packages free of side-effects. I think I need to read up on the rules of what makes a module side-effect free or not. 

For one, we'd have to mark the files you've called out as having side effects, and I suspect that there are probably more - e.g.  https://github.com/react-dnd/react-dnd/blob/c5a9abc7b11ab3bce6ac47af031d2f2970ca67a6/packages/core/react-dnd/src/common/DndProvider.tsx#L70

I wonder if there's a way to verify the sideEffect declaration to make sure tree shaking doesn't drop anything we're supposed to keep.

maclockard:
@darthtrevino I would definitely read up on the rules! Its a bit confusing since what webpack has defined as a "side effect" definitely doesn't match the colloquial understanding of what a side effect is. https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free does a good job with some examples that demonstrate this.

For the example you linked, while there is a value being computed on import, it doesn't count as a side effect using webpack's definition. Webpack consider's things to be side effects if they are executed on module import _and_ outside of an import lineage. Since `instanceSymbol` is used by `createSingletonDndContext` its a part of an import lineage and would not be shaken as long as `createSingletonDndContext` is used.

If it were to look like this instead:
```
let instanceSymbol;
instanceSymbol = Symbol.for('__REACT_DND_CONTEXT_INSTANCE__');
```

`instanceSymbol` would be `undefined` after tree shaking since the statement `instanceSymbol = Symbol.for('__REACT_DND_CONTEXT_INSTANCE__')` is _not_ a part of an import lineage and is only executed on import.

The examples I linked are in fact side effects, but from what I can tell, they are just used during dev so would be safe to drop in production. So there is no harm not declaring them, and we would get even better shaking performance letting them be dropped.

maclockard:
A slightly more involved example of a side effect using webpack's definition:

```js
const myDict = {"a": "junk"};
myDict["a"] = "myVal";

function printMyVal() {
  console.log(myDict["a"]);
}
```

If this module is tree shaken, since only the `myDict` variable declaration is a part of an import lineage, calling `printMyVal()` would print `"junk"`.

`myDict["a"] = "myVal";` would be dropped by the shaking process.

An alternative _side effect free_ version could look like this:

```js
const myDict = () => {
  const dict =  {"a": "junk"};
  dict["a"] = "myVal";
  return dict;
}();

function printMyVal() {
  console.log(myDict["a"]);
}
```

Even though we might normally think of executing a function on import as a side effect, webpack does _not_ consider it to be one since executing it is a part of `myDict`'s import lineage. Calling `printMyVal()` here would print `"myVal"`.

However, if the function generating `myDict` were to mutate some other value:

```js
const otherDict = {"b": "junk"}

const myDict = () => {
  const dict =  {"a": "junk"};
  dict["a"] = "myVal";
  otherDict["b"] = "otherVal";
  return dict;
}();

function printMyOtherVal() {
  console.log(otherDict["b"]);
}
```

calling `printMyOtherVal()` would print `"junk"` if `myDict` is unsued and tree shaken, and would print `"otherVal"` if `myDict` is used somewhere else. This is a particularly insidious version of a side effect and can be difficult to track down.

PR #1610, opened 2019-11-22
fix: filemetadata available on dragstart

LeopoldLerch:
Hi
This should fix the issue #584 . 

I changed it so that when a file is dragged into the browser, it will create the nativesource together with the file informations. Before it simply created placeholders telling that browser wouldn´t allow it. That was only partially true. The browser allows to acces the file´s metadata like name, size, ... . However, it allows not to access its (binary) data. The binary data is available after the ondrop-event. Therefore on the ondrop event the nativedragsoure will be updated again with new DataTransfer-object.

However, this happens only on DragEnter into the browser and also only for the File-Nativetype. All the other ones stay the same. I was not really able to test them out correctly and the security that prevents from accessing the file´s binary data might effect here more, as the Link and escpecially the text don´t have any metadata at all.

I needed this for a rebuild of the "DropZone" with react-dnd. And the only issue was, that I was not able to react on the file´s metainfo (eg. filetype) prior to droping it (eg. marking the dropzone "green" when it is ok).

Pls have a look at it.

Regards
Leo


LeopoldLerch:
Do you have an idea when this will be available via npm?

darthtrevino:
Hopefully this weekend, I have some things to get done at home today to prepare for the Holiday next week. I'll try to cut a release tonight

PR #1631, opened 2019-11-26
Patch TouchBackend to solve e.targetTouches bug in iOS 13

### LINKED ISSUE #1585: Can't drag down on iOS 13
**Describe the bug**
While iOS upgrade to 13, our dnd sortable list component can drag up but can't drag down. All it worked in iOS < 12. Is it a bug in iOS 13 ? Have any idea?

- OS: iOS
- Browser: chrome, safari
- Version 78.0.3904.67

zhusee2:
Based on the discussions and findings over #1585, I believe Safari in iOS 13 can somehow lose `e.targetTouches` after a few renders.

As a workaround I set up a private property in `TouchBackend` to “cache” a valid target `Touch` from `touchstart` event, then use it as a fallback in `touchmove` event handlers when it fails to get client offset from an empty `e.targetTouches` list.

And this should fix #1585.


darthtrevino:
Thanks for the PR! Have you verified that this fix works on an iOS device?

zhusee2:
Yes, I've tried it locally in my project with an iPad running iOS 13.2.3. It does seem working properly.

PR #1733, opened 2019-12-17
Example was using hooks instead of class

jwallet:


darthtrevino:
Thanks for the PR! The decorator-based examples aren't necessarily designed to be free of Function Components or low-level react hooks, but are meant to use the decorator-based API of react-dnd instead of the hooks API. Unless there's a good reason for making this package 100% hook-free, I'm not sure that it buys much.

Issue #1041, opened 2018-05-30
Hover and drop not being called when a DragLayer is used.


joaquindk:
When I use the DragLayer HOC to customise the drag preview, hovering over a drop target doesn't work if the drag preview is placed directly under the mouse pointer. However if the drag preview is placed adjacent to the pointer everything works as expected.

So basically if I define a DragLayer like this...

```
const dragLayerConnector = monitor => ({
  isDragging: monitor.isDragging(),
  itemType: monitor.getItemType(),
  item: monitor.getItem(),
  currentOffset: monitor.getSourceClientOffset(),
  initialClientOffset: monitor.getInitialClientOffset(),
  initialSourceClientOffset: monitor.getInitialSourceClientOffset(),
});

class MyDragLayer extends Component {
  getStyles() {
    const { currentOffset } = this.props;

    if (!currentOffset) {
      return {
        display: 'none',
      };
    }
    const { x, y } = currentOffset;
    
    return {
      transform: `translate(${x}px, ${y}px)`,
      position: 'absolute',
      width: '200px',
    };
  }

  render() {
    const {
      isDragging,
      itemType,
      item,
      currentOffset,
      initialClientOffset,
      initialSourceClientOffset,
      ...other
    } = this.props;

    if (!isDragging) {
      return null;
    }

    if (itemType !== Types.MyItem) {
      return null;
    }

    if (!item) {
      return null;
    }

    const style = this.getStyles();

    return (
      <div style={style}>
        Preview test
      </div>
    );
  }
}

export default DragLayer(dragLayerConnector)(MyDragLayer);

```

..if my drag source is wider than 200px and I begin to drag from beyond the 200px that the drag layer draws, everything works as expected. However if I begin to drag from within the 200px width, where the draglayer preview is rendered, dragging is automatically stopped as soon as the drag layer renders the first time.

I've managed to work around this by leaving the preview drawn by the drag layer one pixel under my pointer, using the following solution inside `getStyles`, however it would be nice if this wasn't necessary:

```
getStyles() {
    const { currentOffset, initialClientOffset, initialSourceClientOffset } = this.props;

    if (!currentOffset || !initialClientOffset || !initialSourceClientOffset) {
      return {
        display: 'none',
      };
    }
    const { x, y } = currentOffset;
    const { y: yInitPointer } = initialClientOffset;
    const { y: yInitSource } = initialSourceClientOffset;

    return {
[...]

PR #1734, opened 2019-12-18
Chrome DragLayer bug fix when sometimes dragend firing immediately after dragstart

newsiberian:
Hi, we faced w/ a bug w/ DragLayer. Something similar was described [here](https://stackoverflow.com/questions/14203734/dragend-dragenter-and-dragleave-firing-off-immediately-when-i-drag). I know, your Sendbox example works fine everywhere, but in our case DragLayer is broken. It fires dragend just after you pickup the item. Maybe because we are using it within MUI Dialog component (modal), I don't know. This PR fix that chrome behavior.

This could be related w/ #1041


darthtrevino:
@newsiberian I'd be glad to include this change - can you set up a small example so I can verify the issue? I'd like to include it in one of the "extra" examples so that we can make sure this your desired behavior doesn't break in future builds, and I'd like to eventually get it under test automation.

PR #2280, opened 2020-04-21
make handleTopDragStart public for extension

th3fallen:
Make handleTopDragStart public for extension for additional modifier key support.

like... 

```
import HTML5Backend from 'react-dnd-html5-backend/lib/cjb/HTML5Backend';
import { TouchBackend } from 'react-dnd-touch-backend';

import { useTouchBackend } from 'shared/util/dnd';

class HTML5Modifiers extends HTML5Backend {
  handleTopDragStart(e) {
    super.handleTopDragStart(e);

    const doCopy = e.altKey || e.ctrlKey || e.metaKey;
    this.wiwDropEffect = doCopy ? 'copy' : 'move';
  }
}

class TouchModifiers extends TouchBackend {
  handleTopMoveStart(e) {
    super.handleTopMoveStart(e);

    const doCopy = e.altKey || e.ctrlKey || e.metaKey;
    this.wiwDropEffect = doCopy ? 'copy' : 'move';
  }
}

export default function createDndBackend(manager) {
  if (useTouchBackend()) {
    return new TouchModifiers(manager, {
      enableMouseEvents: true,
      enableTouchEvents: false,
    });
  } else {
    return new HTML5Modifiers(manager);
  }
}
```


darthtrevino:
If that one is made public it would make sense to make all the handlers public

PR #2423, opened 2020-05-15
Fix memory leak in drag preview, un-subscriber not called

Calyhre:
**TL;DR**: We discovered a memory leak inside `react-dnd` drag previews logic, and it's pretty straight forward to fix. The preview nodes end up never being cleaned, if the preview is attached with the `useDrag` hook.

Here is a reproduction within an example of the documentation -> https://codesandbox.io/s/suspicious-sky-ytgu2

---

When using [`useDrag`](https://github.com/react-dnd/react-dnd/blob/master/packages/core/react-dnd/src/hooks/useDrag.ts), `react-dnd` instantiate or retrieve a memoized [`SourceConnector`](https://github.com/react-dnd/react-dnd/blob/master/packages/core/react-dnd/src/common/SourceConnector.ts). The hook then returns some state attributes and 2 refs, one for the drag source and the other for the preview.

Both are function refs, meaning they will be called by React with the ref at mounting, update, and un-mounting. Called with `null` will tell the connector that the node is being unmounted.

Those two functions are visible here: [`dragSource` & `dragPreview`](https://github.com/react-dnd/react-dnd/blob/master/packages/core/react-dnd/src/common/SourceConnector.ts#L17-L39).

Let see how `dragSource` work as example, as they are both similar:

```typescript
dragSource: (
  node: Element | React.ReactElement | React.Ref<any>,
  options?: DragSourceOptions,
) => {
  this.clearDragSource()
  this.dragSourceOptions = options || null
  if (isRef(node)) {
    this.dragSourceRef = node as React.RefObject<any>
  } else {
    this.dragSourceNode = node
  }
  this.reconnectDragSource()
},
```

Pretty simple: It clears some stuff, assign the new values to different internal variables, then call reconnect. <mark>**This specific order will be very important later.**</mark>

Now let's see [`reconnectDragSource`](https://github.com/react-dnd/react-dnd/blob/11add3368027907753ca68bc59b2a954d1eb32c2/packages/core/react-dnd/src/common/SourceConnector.ts#L100):

```typescript
const dragSource = this.dragSource;
// if nothing has changed then don't resubscribe
const didChange =
  this.didHandlerIdChange() ||
  this.didConnectedDragSourceChange() ||
  this.didDragSourceOptionsChange();

if (didChange) {
  this.disconnectDragSource();
}

if (!this.handlerId) {
  return;
}
if (!dragSource) {
  this.lastConnectedDragSource = dragSource;
  return;
}

if (didChange) {
  this.lastConnectedHandlerId = this.handlerId;
  this.lastConnectedDragSource = dragSource;
  this.lastConnectedDragSourceOptions = this.dragSourceOptions;
  this.dragSourceUnsubscribe = this.backend.connectDragSource(
    this.handlerId,
    dragSource,
    this.dragSourceOptions
  );
}
```

First line is a bit confusing at first as `dragSource` does not exists anywhere. Instead, it's a getter used to DRY the access of one or another internal values, [as seen here](https://github.com/react-dnd/react-dnd/blob/11add3368027907753ca68bc59b2a954d1eb32c2/packages/core/react-dnd/src/common/SourceConnector.ts#L197-L201).

Then it compute a `didChange` flag comparing old & new values, and finally perform internal cleaning based on it. The `connectDragSource` method returns a method to call at cleaning time. You can see the usage in [`disconnectDragSource`](https://github.com/react-dnd/react-dnd/blob/11add3368027907753ca68bc59b2a954d1eb32c2/packages/core/react-dnd/src/common/SourceConnector.ts#L181-L186).

And that's where what I think should be the same logic for the preview fails to be. In [`reconnectDragPreview`](https://github.com/react-dnd/react-dnd/blob/11add3368027907753ca68bc59b2a954d1eb32c2/packages/core/react-dnd/src/common/SourceConnector.ts#L132-L153), it uses another getter, another `didChange` flag, and then some cleaning logic. But there is a catch, the disconnect is inside a block that checks that `dragPreview`, [`dragPreviewNode` or  `dragPreviewRef` behind the getter](https://github.com/react-dnd/react-dnd/blob/11add3368027907753ca68bc59b2a954d1eb32c2/packages/core/react-dnd/src/common/SourceConnector.ts#L203-L208), is truthy. Or, in the exact same execution flow, we saw in the `dragSource` & `dragView` code above that those variables are wiped before calling this reconnect logic and so this whole branch of code is never called. 

The fix is pretty simple, I duplicate the logic from the drag source, and added a few specs to ensure everything is fine. 

Let me know what you think :)


darthtrevino:
@Calyhre I added a regression test to the docs examples with your memory leak example. It looks like the memory leak is still occurring. You can verify this if you fire up the docsite and go to `Examples -> Other Cases -> Previews Memory Leak`

Calyhre:
Hey @darthtrevino, thanks for the regression test! I found the reason why: my fix was half of the issue in the example page, [`DragPreviewImage`](https://github.com/react-dnd/react-dnd/blob/master/packages/core/react-dnd/src/common/DragPreviewImage.tsx) is indeed introducing its own memory leak too. 

The ref callback function, locally called `connect`, is called on `img.onload` but never called on unmount.

I moved all the connection logic into a `useEffect` with no dependencies, and call the ref callback function with `null` on the returned cleanup function, implementing the same behavior as a classic `React` ref. 

Tell me what you think :)

darthtrevino:
Awesome, it may make sense to do additional profiling work at some point to make sure element references aren't getting out of hand, but this looks great. Thanks @Calyhre

PR #2491, opened 2020-05-26
Resolve circular dependencies in react-dnd

### LINKED ISSUE #2441: Circular dependencies in react-dnd limit usage possibilities
Using react-dnd in two contexts:

1. With a bundler that supports circular dependencies.
2. With a bundler that does not support circular dependencies.

I noticed that there are at least 3 cycles in the dependency graph of the react-dnd package (there may be more, but these are the ones I noticed), and as such, we can't easily use react-dnd with bundler "2" (the bundler throws an error as soon as it detects a cycle):

- [index.ts](https://github.com/react-dnd/react-dnd/blob/1d5d3b0db1826beb9c36dea0c7df8dafc662aeb4/packages/core/react-dnd/src/index.ts) depends on:
  - [decorators/index.ts](https://github.com/react-dnd/react-dnd/blob/1d5d3b0db1826beb9c36dea0c7df8dafc662aeb4/packages/core/react-dnd/src/decorators/index.ts), which depends on each of:
    - [decorators/DragLayer.ts](https://github.com/react-dnd/react-dnd/blob/1d5d3b0db1826beb9c36dea0c7df8dafc662aeb4/packages/core/react-dnd/src/decorators/DragLayer.tsx)
    - [decorators/DragSource.ts](https://github.com/react-dnd/react-dnd/blob/1d5d3b0db1826beb9c36dea0c7df8dafc662aeb4/packages/core/react-dnd/src/decorators/DragSource.tsx)
    - [decorators/DropTarget.ts](https://github.com/react-dnd/react-dnd/blob/1d5d3b0db1826beb9c36dea0c7df8dafc662aeb4/packages/core/react-dnd/src/decorators/DropTarget.tsx)
  - Each of which in turn depend on the top-level index.ts, either directly or via [decorators/decorateHandler.tsx](https://github.com/react-dnd/react-dnd/blob/1d5d3b0db1826beb9c36dea0c7df8dafc662aeb4/packages/core/react-dnd/src/decorators/decorateHandler.tsx), creating a cycle.

Workarounds:

1. Make a pre-built bundle containing `react-dnd` using a bundler that can handle the cycles, and then feed *that* bundle into "bundler 2".
2. Stop using "bundler 2"; use "bundler 1" instead (sadly, not viable in all contexts).
3. Fork react-dnd or otherwise patch it at build time to remove the cycles.

Ideally though, if the cycles could be removed here, then react-dnd would work with the largest possible range of bundlers. It seems the main reason `DragLayer`, `DragSource` and `DropTarget` are creating a cycle is because they want access to `DndContext` via the top-level `index.ts`, but it seems that they could just get it from [common/DndContext.ts](https://github.com/react-dnd/react-dnd/blob/1d5d3b0db1826beb9c36dea0c7df8dafc662aeb4/packages/core/react-dnd/src/common/DndContext.ts) instead.

darthtrevino:
Fixes #2441


wincent:
Thanks @darthtrevino!

PR #2499, opened 2020-05-26
TouchBackend - impossible to drop item 

tony-bondarenko:
https://github.com/react-dnd/react-dnd/blob/55957e32b57998de02f3ef9621a1d4d2c47546d1/packages/core/touch-backend/src/TouchBackendImpl.ts#L531
this.targetNodes is a Map. It's not possible to iterate over map keys using for (... in ...). As a result, orderedDragOverTargetIds array is always empty.
Because of that drop callback from useDrop is never being called (when an item is dropped to drop zone).
Changing this line to 
`for (const targetId of this.targetNodes.keys()) {`
do the trick and drop became possible after that.


darthtrevino:
Thanks @tony-bondarenko , we'll get this released today

PR #2507, opened 2020-05-27
Fix touch backend bug in _getDropTargetId

tony-bondarenko:
This is a fix of an issue missed in yesterday`s touch backend fix. Now only the first node in a map can be found.


darthtrevino:
Derp on my part, thanks @tony-bondarenko

PR #2756, opened 2020-09-23
feat(touch-backend): add option to begin drag after delayed start

aomader:
This essentially allows to trigger the start of a drag operation without requiring the user to move after he has pressed/touched for a certain amount of time.


darthtrevino:
This looks reasonable, I may change the config name on merge

darthtrevino:
@aomader Apologies for getting to this late. It looks like the touch backend has configuration options for supporting touch delays (e.g. delay, delayTouchStart, and delayMouseStart). IIRC, if these are set to values greater than zero you can support delays on dragging. And I believe the `delay` attribute will configure `delayTouchStart` and `delayMouseStart` at once. 

Does this PR expand upon this in some way? Were those configuration values insufficient?

aomader:
@darthtrevino The current configuration only ensures that a drag operation is started if some time has passed between start and first movement. The important bit is that the drag operation is started **after** the movement. I.e., just pressing/holding something is not sufficient.
This is why I added this option, which allows to specify the additional behaviour to trigger the drag after the `delay*` amount without requiring an extra movement. I would argue that this is actually the intuitive behaviour. If I touch down on something and wait for the specified time, I would want to see a drag indication (which is only possible to visualize after the drag operation has started) when that time has passed, rather than when I start moving.

PR #2830, opened 2020-11-01
Add 'id' property to 'DragObjectWithType'

hassaans:
Right now if we try to use the 'id' property of a dropped item, we get this error:

```
Failed to compile.

/Users/.../index.tsx
TypeScript error in /Users/.../index.tsx(26,35):
Property 'id' does not exist on type 'DragObjectWithType'.  TS2339

    23 |const [{ canDrop, isOver }, drop] = useDrop({
    24 |  accept: 'text',
    25 |  drop: (item, monitor) => {
  > 26 |         handleChangePosition(item.id)
       |                                   ^
    27 |  },
    28 |  collect: monitor => ({
    29 |    isOver: monitor.isOver(),
    30 |    canDrop: monitor.canDrop(),
    31 |  }),
    32 |});

```

This change will allow us to use the 'id' property in `useDrop`. Plz, lemme know if I need to make further changes to get this approved.

### DIFF
--- a/packages/react-dnd/src/interfaces/hooksApi.ts
+++ b/packages/react-dnd/src/interfaces/hooksApi.ts
@@ -126,4 +126,5 @@ export interface DropTargetHookSpec<DragObject, DropResult, CollectedProps> {
 
 export interface DragObjectWithType {
 	type: SourceType
+	id?: string
 }


hassaans:
Test related to version is failing. Do I need to change the destination branch?

darthtrevino:
You would need to run `yarn version check --interactive` to create a semver impact document. I'll take a closer look at this tomorrow sometime. If the id property was already present, fixing the typings would be patch

darthtrevino:
After looking at this some more, I'm not sure if this makes sense. You can provide your own type with an `id` field that extend `DragObjectWithType`. The first generic argument of `useDrag` can be your custom type.

hassaans:
That's weird. I'm able to add `id` to `useDrag`'s first argument, even if I don't change `DragObjectWithType`. It should be a part of it though as it's recommended in the docs as well.

Screen Shot 2020-11-14 at 10 38 00 AM

PR #2848, opened 2020-11-10
Add document owner property to touch backend

jcampbell05:
In certain projects such as the one I am working on elements are encapsulated so that they are in their own sandbox. In modern browsers you would do this with the ShadowDOM but since we have to support IE 11 we cannot use that API.

We could use a Polyfill but as we are an SDK that would add unneeded weight. Instead we encapsulate by mounting our element within an iframe. 

Unfortunately that breaks some assumptions this library makes. The element is now owned by the document of the iframe and not the root document object meaning events can only be listened to from the document object for the iframe.

The library assumes that it will always be able to get these events from the global document for the window. In this PR I provide an ownerDocument option which allows the consuming project to provide it's own document object for the touch backend to attach its events to.

So within the encapsulated component you can pass in a ref to a `div` element's ownerDocument to the backend, like so:

```js
const collection = () => {
    const container = useRef(null)
    return <div ref={container}>
         <DNDContext options={{ ownerDocument: container.current.ownerDocument} }/>
    <./div>
}
```

Similar changes will need to be made for the HTML5 backend but I wanted to keep this PR focused on one backend.

### DIFF
--- a/packages/backend-touch/package.json
+++ b/packages/backend-touch/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-dnd-touch-backend",
-  "version": "11.1.3",
+  "version": "11.1.4",
   "description": "Touch backend for react-dnd",
   "main": "dist/cjs/index.js",
   "module": "dist/esm/index.js",
--- a/packages/backend-touch/src/OptionsReader.ts
+++ b/packages/backend-touch/src/OptionsReader.ts
@@ -5,6 +5,7 @@ import {
 } from './interfaces'
 
 export class OptionsReader implements TouchBackendOptions {
+	public ownerDocument: Document | null = null
 	public enableTouchEvents = true
 	public enableMouseEvents = false
 	public enableKeyboardEvents = false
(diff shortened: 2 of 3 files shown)


darthtrevino:
LGTM, does the HTML5 backend need a similar configuration?

jcampbell05:
> LGTM, does the HTML5 backend need a similar configuration?

Yes I believe it will.

jcampbell05:
> LGTM, does the HTML5 backend need a similar configuration?
In CI I am seeing a message about "some workspaces had changes but had no instructions on how they should be released" do you know how I can resolve this ?

darthtrevino:
Try running ‘yarn version check —interactive’ to create a semver impact document

jcampbell05:
> Try running ‘yarn version check —interactive’ to create a semver impact document

Just added a patch release strategy hopefully that helps

darthtrevino:
@jcampbell05 Can you grant me write access on your branch?

jcampbell05:
> Can you grant me write access on your branch?
> @jcampbell05 Can you grant me write access on your branch?

You will now have access

darthtrevino:
Hmm, I'm having trouble pushing. Can you look on the right side of the PR? There should be a checkbox that says "allow edits by maintainers" or something to that effect.

jcampbell05:
> Hmm, I'm having trouble pushing. Can you look on the right side of the PR? There should be a checkbox that says "allow edits by maintainers" or something to that effect.

Doesn't show up unfortunately, I sent you an invite to my repo instead can you accept that and see

darthtrevino:
Merged with #2859

PR #2919, opened 2020-12-16
Use event processing techniques to improve performance

jcampbell05:
During testing of our SDK it appears that some browsers can emit screen tearing on lower end devices. Potentially this is sue to slow event handling during screen renders.

During profiling I found some event handlers could take as much as 3 seconds to fire on a VM with very low memory and low core counts. Which goes over our frame budget.

I've tried the following techniques which seem to result in event handlers taking 20 - 50ms to fire which should allow for more timely updates during repaints.

- We throttle mouse/touch events to only handle them every 50 milliseconds as most move updates only translate to a very minor move.
- Instead of requesting an animation update on every move we start an animation frame update loop when we detect the first move which means we will get regular frame updates when the browser is ready to handle them.

I also attempted to make all event handlers passive but it appears in a few places events are cancelled. Potentially this is another optimisation that can be done if needed.

### DIFF
--- a/.yarn/versions/01c379ac.yml
+++ b/.yarn/versions/01c379ac.yml
@@ -1,6 +1,3 @@
-releases:
-  react-dnd-html5-backend: patch
-
 declined:
   - react-dnd-documentation
   - react-dnd-examples-decorators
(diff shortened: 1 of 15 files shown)


jcampbell05:
Wrong repo

Issue #2928, opened 2020-12-22
Html5 backend should support 'text/html' type


maxired:
**Is your feature request related to a problem? Please describe.**
I am using react-dnd to handle file dropping inside my app.
Some people are trying to drag and drog image from other tab form this browser. we currently receive the event with the data type corresponding to NativeTypes.URL. But the data transfer object I receive also has some data with type  `text/html`. I would like to be able to retrieve it.

**Describe the solution you'd like**
I want to be able to retrieve data with type `text/html`

darthtrevino:
Thanks for the PR!

PR #2949, opened 2021-01-05
add html handling in native types

maxired:
This PR solves https://github.com/react-dnd/react-dnd/issues/2928 and allows to drag and drop HTML elements such as images from another tab.

This PR does not implements any custom logic to deal with the HTML data in itself.

### DIFF
--- a/.yarn/versions/88ed6e63.yml
+++ b/.yarn/versions/88ed6e63.yml
@@ -1,0 +1,5 @@
+releases:
+  react-dnd-documentation: patch
+  react-dnd-examples-decorators: patch
+  react-dnd-examples-hooks: patch
+  react-dnd-html5-backend: patch
--- a/packages/backend-html5/src/HTML5BackendImpl.ts
+++ b/packages/backend-html5/src/HTML5BackendImpl.ts
@@ -649,6 +649,7 @@ export class HTML5BackendImpl implements Backend {
 		this.dropTargetIds = []
 
 		if (this.isDraggingNativeItem()) {
+			e.preventDefault();
 			this.currentNativeSource?.loadDataTransfer(e.dataTransfer)
 		}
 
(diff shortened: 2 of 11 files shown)


maxired:
@darthtrevino do you have any opinion about this ?

darthtrevino:
I apologize, I've been a beet slammed at work. Just taking a quick look - Is the preventDefault() necessary()? At some point we removed some of that because we didn't want to interfere with custom handling of native events. Also, can you put together an example on the docsite using this new type? You can base it off of the existing native types example or just extend it if you prefer

maxired:
Thanks for your answer. Yes I will try to setup an example.

Without the preventDefault, the browser opens the content in a new tab in that case. I am not very familiar with this code base, so maybe it can be put somewhere else? As a side not, I did not saw side effects on other handlers.

maxired:
@darthtrevino I've just added an example, please let me know if this looks good to you.

maxired:
@darthtrevino sorry to bother you. What is the status on this one ?

darthtrevino:
Hey Maxence,
Again, apologies for the delay. I went ahead and merged it, I'll work on
stabilizing the main branch and preparing for a release.

inukshuk:
This PR raises a problem: dragged images, URLs, and links etc. typically have multiple data transfer items with type `text/html` and `text/uri-list` and even `text/plain`. Right now it seems that [matchNativeItemType](https://github.com/react-dnd/react-dnd/blob/main/packages/backend-html5/src/NativeDragSources/index.ts#L13) returns just the first match: that means for example that dragged images or URLs are now only matched as type `HTML`, because it was added higher up in the native types config. For this reason, this PR makes it impossible (I believe?) to register drop targets for URLs -- unless you register for type HTML and then parse out the URL from the HTML data.

PR #2733, opened 2020-08-21
Testing.md: React Testing Library & React DnD Chessboard Example

ryota-murakami:
[Original PR](https://github.com/react-dnd/react-dnd/pull/2729) was failed CI at `lint:spelling`, I'm sorry about it.
I fixed it and then submit PR again.

Please review whenever you have time. 🙇‍♂️

Thank you maintaining continually awesome library!

PR #3025, opened 2021-02-11
Migrate Chess Example Tests from PR

darthtrevino:
Thanks @ryota-murakami, these tests are much clearer to read and author than what we have using the `react-dnd-test-backend` and `react-dnd-test-utils` packages. I love how easy it is to declare drag/drop interactions using the testing-library API. 

Adapted tests from #2733


ryota-murakami:
@darthtrevino That's great! 
I'm glad to improved official testing example as a result.

Thank you tell me about it!

PR #3048, opened 2021-02-20
Resolve codesandbox React errors and warnings

192kb:


192kb:
Was edited directly on codesandbox, so styling was ruined a bit. 
In overall, it just requires React to be imported for JSX tags, I guess

darthtrevino:
@192kb I'd prefer to use React 17 idioms for the examples, and this PR reverts a lot of our intentional stylistic choices - thank you for the heads up though, I'll dig into the examples config

darthtrevino:
Thanks again for the report! I updated the boilerplate to use a newer version of react-scripts that supports React 17 changes.

PR #3114, opened 2021-03-04
useDrag() Hook Changes

### LINKED ISSUE #3110: monitor.getItem() always returning null
**Describe the bug**
When I call `monitor.getItem()` inside `canDrag()` it always returns `null`. 

**Reproduction**

[Live Reproduction](https://codesandbox.io/s/blue-feather-413m0?file=/src/Box.tsx)

1. Drag around one of the boxes
2. Check the console to see the value of `monitor.getItem()` (it's always `null`)

**Expected behavior**
I expect the result to be either the value of `item` or whatever is returned by `begin()`. This is particularly useful in combination with `begin()` as some drag events need to be suppressed based on its result. 

- OS: macOS
- Browser: chrome
- Version: 88.0.4324.192 (Official Build) (x86_64)

darthtrevino:
The `useDrag` hook has two modes of operation: building out return items with `spec.item` and with `spec.begin()`. When you use `spec.begin()`, objects are expected to be of the type in `spec.item.type`, which muddles concepts and can lead to some confusion. The deepest change here is separating the concerns of **draggable type** from **draggable items**.

This change is a major semver change that refactors the `useDrag()` ergonomics.
* `spec.type` is required. This should be the same value that was in `spec.item.type`. This simplifies use cases that _only_ use drag-types and no further drag-object details.
* `spec.item` can be an object or function. The object does not need to contain the `type` property. The function may return an item object or `null`. `null` cancels the drag operation. The function form of `item` replaces `begin`, which is no longer present.
* end() provides the dragObject and dropResult as arguments. (e.g. `end(item, result, monitor)`)

The typings for monitors has been tightened. The interfaces for DragSourceMonitor and DropTargetMonitor now have type parameters for the DragObject and DropResult which default to unknown. Furthermore the getItem() and getResult() methods can be parameterized with the type expected (default is monitor typings). 

Finally, `dnd-core` has been updated so that if a dragHandler returns `null` or `undefined`, the system will cancel the drag operation. This allows clients using `useDrag({ item: () => condition ? null: validItem })` to cancel drag operations with client-side logic.

Bugfixes include correcting liveness issues with hooks that were evident in the Dustbin stress test and fixing infinite render loops from shallowEqual being used when deepEqual is necessary.

Fixes #3110


Mangatt:
@darthtrevino It might be good idea to put deprecated warning when someone tries to use`begin` function. Right now it's bit difficult to trace what went wrong after update, especially when this is not listed as breaking change in changelog.

darthtrevino:
Right - I struggled a bit with whether to deprecate `begin` or just nuke it with this release. At the very least we can add a friendly developer message if someone defines `spec.begin` to let them know to switch to `spec.item`

darthtrevino:
I'm adding a developer message here - https://github.com/react-dnd/react-dnd/pull/3151

PR #3118, opened 2021-03-05
Implements performance improvments

jcampbell05:
During implementation in our SDK we found a few ways we can improve the performance and UX of the library. There will be a few changes that may need to be rolled back in this PR.

I had to move the ownerDocument change into the context object since it breaks when in the options area.


jcampbell05:
Closed due to condensed PR 😀

PR #3145, opened 2021-03-17
Change isActive to isOver

cursorsdottsx:
In the code shown in this markdown file, I couldn't find the declaration for `isActive`, and I believe you meant `isOver`?


darthtrevino:
It looks like this one was caught by another reviewer as well. Thanks for the PR!

Issue #3138, opened 2021-03-12
Can't Drag Native Files into iframe


nshenkman:
Hi there!

**Describe the bug**
In my testing I was unable to drag native files into an iframe. Essentially the behavior I saw was that the isOver/canDrop properties from useDrop would return true and then immediately false making it impossible to actually to drop the files. 

In the example below I essentially merged the React DnD iframe example and the native file example into one sandbox. If you try dragging a file over you'll notice that the isOver/canDrop variables becomes true and then immediately false as I drag around the drop area.

**Reproduction**

[Live Reproduction](https://codesandbox.io/s/eloquent-sanderson-zkb8t)

1. Go to https://codesandbox.io/s/eloquent-sanderson-zkb8t
2. drag a file into the drop area
3. Notice flickering of dragging state and hover state (due to isOver/canDrop switching between true and false)

**Expected behavior**
Should behave the same way it does outside of iframe 

https://codesandbox.io/s/github/react-dnd/react-dnd/tree/gh-pages/examples_hooks_ts/06-other/native-files

**Additional context**
I debugged the issue and identified that this is happening due to `isNodeInDocument` check inside of the HTML5Backend implementation. In this check, [the code](https://github.com/react-dnd/react-dnd/blob/main/packages/backend-html5/src/HTML5BackendImpl.ts#L320) is defined as 
 
```
private isNodeInDocument = (node: Node | null | undefined): boolean => {
	// Check the node either in the main document or in the current context
	return Boolean(
		node &&
		this.document &&
		this.document.body &&
		document.body.contains(node),
	)
}
```

The issue stems from the last line `document.body.contains(node)` because it is not using the context document but rather the window one. I believe the fix would be as simple as updating it to be `this.document.body.contains(node)` but wanted to see if there was a specific reason it had been implemented this way before submitting a PR. 

I confirmed that this fixed the issue by overriding the window document.body.contains function to check both the parent and the specific child iframe (in no way is this a good solution but rather was just to show that this would solve the issue)
[Example](https://nhslq.csb.app/)
[Code](https://codesandbox.io/s/new-tdd-nhslq?file=/src/Container.tsx) for some reason the drag and drop functionality wasn't working within codesandbox but did work in the popped out example above

nshenkman:
Can anyone take a look?

PR #3181, opened 2021-04-23
Fix drop operations in iframes/child windows

eramdam:
We were running into the same bug as in #3138 on our app at @frontapp and I confirmed that the diff fixes the issue for us using `patch-package`, so I'm opening a PR to help others with the same issue 😁 

Fixes https://github.com/react-dnd/react-dnd/issues/3138


eramdam:
Looks like https://github.com/react-dnd/react-dnd/pull/3271 fixes that and was just merged 😁

Issue #3157, opened 2021-03-24
Textarea drop behavior HTML5Backend


ShaMan123:
**Describe the bug**
Native textarea html behavior is prevented (I guess because of `e.preventDefault()`) causing dropping text to be awkward.

**Reproduction**
[Live Reproduction](https://codesandbox.io/s/old-framework-lizc2)

1. Use `Html5Backend`
2. Try dropping text into a textarea
3. Switch to `TouchBackend`
4. same as 2

**Expected behavior**
I expected `HTML5Backend` to at least show the textarea's caret/cursor and perhaps leave the text insertion as opt in while providing relevant data so the text can be inserted correctly.

**Desktop (please complete the following information):**

- OS: [e.g. iOS] Windows
- Browser [e.g. chrome, safari] chrome
- Version [e.g. 22] Latest

**Additional context**

vitpekarek:
So there is no temporary solution I guess? Is there any estimation when fix get released? Thank you

ahoisl:
As a workaround, just add `onDragOver` event handler on the text areas that stop event propagation:
`<textarea onDragOver={evt => evt.stopPropagation()} ... />`

Issue #3179, opened 2021-04-18
Dropping a link follows it in Firefox.


galenhuntington:
**Describe the bug**

I have an `<a href="">` link element which I want to make also draggable.  I had wrapped the link in a `<span ref={drag}>` to be sure.  Previously this was fine; clicking it works as a link, but it also instead can be dragged.  However, after upgrading `react-dnd-html5-backend` to the latest 14.0.0, this no longer works; as soon as it is dropped over a qualified element, the link is immediately followed.  Dropping it elsewhere does not cause this.  I only see the wrong behavior in Firefox, not Chrome.

I have traced it back to #2949.  If I restore the `e.preventDefault()` in `handleTopDropCapture`, it works correctly.  I don't understand the code well enough to diagnose further.  Maybe somehow that `isDraggingNativeItem` has browser differences.

Possibly related, but even the simple example here https://react-dnd.github.io/react-dnd/examples/dustbin/single-target does not work right for me in Firefox.  When I drag into the box, the alert appears _twice_, whereas in Chrome it appears once.  Furthermore, after the second alert, I get an "invariant violation" error in the console.  This might not be the same problem, however.

**Desktop (please complete the following information):**

- OS: Ubuntu Linux
- Browser Firefox 87.0

galenhuntington:
Adding `.preventDefault()` to the drop target's `onDrop` as described in #3188 allowed me to work around this problem.  I don't know if that means it should be considered "fixed".

jgonera:
This looks like a regression. I didn't have this problem on 11.x but it started happening after updating `react-dnd` to 14.0.2.

Seems similar to https://github.com/react-dnd/react-dnd/issues/221 which was fixed a long time ago.

delian1986:
Another workaround is to set manually `draggable=false` to `img` element
`    <div style={{ opacity }} className="footerElement" ref={drag}>
      <img src={data.imgUrl.url ?? ""} alt="" draggable="false" />
    </div>`

istateside:
I think this is happening because the `isDraggingNativeItem` function in HTML5BackendImpl only returns true if the drag source defines its item type as a native drag.

If you mark an element as a drag source with `useDrag`, you define its drag type. However, if an element inside of that drag source could trigger a native drag event (like an `<img>` or `<a>` node), the browser will attach native HTML5 drag types to the `e.dataTransfer` object.

[...]

PR #3188, opened 2021-04-29
Fix/remove html5 native prevent default

maxired:
This PR removes a problemetics preventDefault, and close https://github.com/react-dnd/react-dnd/issues/3157
This is also related to https://github.com/react-dnd/react-dnd/issues/3179 

I also updated the sample to show how a user can call preventDefault manually if need.


darthtrevino:
Hey @maxired ; apologies for taking so long on this. 

I like the idea of this, but it would be breaking behavior and I would get a PR the next day to "fix it" and revert to the old behavior. How about we approach this via an options flag in the HTML5Backend configuration object.

I started a PR here, but I don't have a good test case to verify whether it would work for cases like yours:

https://github.com/react-dnd/react-dnd/pull/3292

ahoisl:
To fix #3157 I guess also the `preventDefault` in the "onDragOver" event handler needs to be removed. See sample https://codesandbox.io/s/old-framework-lizc2

jeremyspiegel:
@darthtrevino would it make sense to skip calling `e.preventDefault()` if the target of the drop is a descendant of a `contentEditable` element? That would still block the "drop and blow away the whole document" action I think you're trying to prevent, but allow dropping text and images into editable sections of the page.

PR #3214, opened 2021-05-24
Remove files that should have been ignored from .git history

valtism:
Currently, the `react-dnd` repo sits at 435MB! I was surprised to see this when I tried to clone the repo.

99% of this size comes from files that have accidentally committed and are now saved in the git history. I have used the shell command: 

```sh
git rev-list --all --objects | \
sed -n $(git rev-list --objects --all | \
cut -f1 -d' ' | \
git cat-file --batch-check | \
grep blob | \
sort -n -k 3 | \
tail -n40 | \
while read hash type size; do
    echo -n "-e s/$hash/$size/p ";
done) | \
sort -n -k1
```

to find what the offending files were.

I then used `git-filter-repo` to remove these files from history. None of the removed files will have an effect on the functionality of the library. I matched the following patterns to remove:

```
--path:
.yarn
examples_ts/00-chessboard/node_modules

--path-regex:
.+\.js\.map$
\.pnp.js$
^main-.+\.js$
^.{40}-.{20}.js$
^.{1}-.{20}.js$
^.{2}-.{20}.js$
^component---.+\.js
```

This has removed all files larger than the `yarn-lock` files and reduces the repo size to 50MB. There could probably be other history to delete.

This is a big git history rewrite, so It will impact all open branches and PRs which will need to be rebased after this is merged. It is not a simple job, but having a repo of such a large size will impact the ability for developers in countries with limited internet, and should not be taken lightly imo.


valtism:
This seems to have breaking changes.

amunrarara:
@valtism Is there any follow-up on this? It's a pretty hefty repo, still, and growing it seems!

valtism:
@amunrarara I was not able to re-write history without causing issues. I think that someone might be able to do it better than me, but you would have to investigate and do the work yourself. 

If you want to check out the repo without history to avoid downloading so much, you can use `git clone --depth 1 https://github.com/react-dnd/react-dnd/` for a shallow clone, where depth is how many commits back you want.

PR #3227, opened 2021-06-08
Add missing "are"

aermler:


Devansh2308 (approved):

froatsnook (commented):
Seems like if we're fixing grammar we might as well just get rid of the whole first part (since it seems to be just repeating the end of the sentence before!). I would suggest:

> Modern React applications have replaced the Higher-Order-Component pattern with hooks. Hooks are a feature of React, introduced in 16.8, that allow for developers to write stateful function components. They are also fantastic for interacting with external stateful systems (\***cough**\* like a Drag-and-Drop engine \***cough**\*).

PR #3261, opened 2021-07-17
Add missing DropTargetMoniter type arguments

equt:
This should allow cases like

https://github.com/react-dnd/react-dnd/blob/e8bd6436548d96f6d6594f763752f424c2e0834b/packages/examples-hooks/src/01-dustbin/copy-or-move/Dustbin.tsx#L40

to have type inference.

### DIFF
--- a/packages/react-dnd/src/hooks/types.ts
+++ b/packages/react-dnd/src/hooks/types.ts
@@ -134,5 +134,7 @@ export interface DropTargetHookSpec<DragObject, DropResult, CollectedProps> {
 	/**
 	 * A function to collect rendering properties
 	 */
-	collect?: (monitor: DropTargetMonitor) => CollectedProps
+	collect?: (
+		monitor: DropTargetMonitor<DragObject, DropResult>,
+	) => CollectedProps
 }


equt:
Wait, is this a bug of GitHub?

darthtrevino:
Oh that's weird, it nuked your branch? I merged some other PRs that were pending. It looks like this is a bot you installed on your fork?

equt:
Yes, I installed [the pull bot](https://github.com/apps/pull) for synchronizing my forks, it basically does a simple fetch and force push I guess.

But the problem here is why the CI is triggered? This is my first pull request in this repository, so it should be blocked. Did you trigger that @darthtrevino?

darthtrevino:
Hey @equt - I'm considering this one. It definitely improves the typings, but it compromises the development ergonomics. A developer would now have to specify 3 generic parameters explicitly when using `useDrag()`. I've tried to make that implicit as possible, which is why I've left those generic args out, but I'm open to reconsidering.

equt:
The actual reason for this pull request is that I found the generic arguments for the `monitor` of `DragSourceHookSpec` is provided

https://github.com/react-dnd/react-dnd/blob/ce9dd69ff577e8e45afd43e95b7ce3ac7a8eac26/packages/react-dnd/src/hooks/types.ts#L85-L87

However the one for the `DropTargetHookSpec` is not

https://github.com/react-dnd/react-dnd/blob/ce9dd69ff577e8e45afd43e95b7ce3ac7a8eac26/packages/react-dnd/src/hooks/types.ts#L137

mifozski:
Why not also provide the arguments to the rest of the `DropTargetHookSpec`'s `monitor` usages? Like `drop`, `hover`, etc.

darthtrevino:
These types have been incorporated in #3300; thanks everyone!

PR #3271, opened 2021-07-30
Fix isNodeInDocument for HTML5Backend when used in iframes

istateside:
The isNodeInDocument function is used to check if a node was detached during a drag event. It checks to make sure `this.document.body` is defined, but then checks if `document.body` contains the given node.

When this library is used within an iframe, `document` in the global context may not be the same as `this.document`. This commit updates the check to use `this.document.body`, which should correctly check against the current context instead of the global document.

### DIFF
--- a/packages/backend-html5/src/HTML5BackendImpl.ts
+++ b/packages/backend-html5/src/HTML5BackendImpl.ts
@@ -323,7 +323,7 @@ export class HTML5BackendImpl implements Backend {
 			node &&
 				this.document &&
 				this.document.body &&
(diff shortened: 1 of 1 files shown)


istateside:
Hey @darthtrevino , any chance I can get a review on this?

darthtrevino:
Looks great, I'll merge on Sunday when I'm back from backpacking

darthtrevino (approved):

istateside:
@darthtrevino sorry for another ping, but could we cut a new release to get this code up for public use?

PR #3290, opened 2021-08-12
Cleanup Connected Items in Unmount

darthtrevino:


packages/react-dnd/src/hooks/useDrop/useDropTargetConnector.ts
```diff
@@ -15,6 +15,7 @@ export function useDropTargetConnector(
 	useIsomorphicLayoutEffect(() => {
 		connector.dropTargetOptions = options || null
 		connector.reconnect()
+		return () => connector.disconnectDropTarget()
 	}, [options])
```
ujaval403, line 19: connector not here in dependency array so clean up not run on connector change

PR #3292, opened 2021-08-12
Add Conditional preventDefault for Native Types

darthtrevino:


ahoisl:
@darthtrevino What would you need to complete this PR? Because this "preventDefault" is causing DnD in our text editor to be broken (from user perspective) as soon as we have a single useDrag/useDrop hook anywhere in our app 😒

ahoisl:
To fix #3157 I guess also the `preventDefault` in the "onDragOver" event handler needs to be removed. See sample https://codesandbox.io/s/old-framework-lizc2

darthtrevino:
After running a clean on the repo, the history got mangled here. I'll need to re-cut this PR

PR #3311, opened 2021-10-05
fix(DndProvider): Add explicit types for children

eps1lon:
We plan to remove implicit `children` from `React.FunctionComponent` and `React.Component`. `DndProvider` currently implements `children` without declaring them in its props interface.

I only found this when running tests in the DefinitelyTyped repository. There may be more problems in this repository but I don't know how to type-check this repository.


darthtrevino:
Thank you for the heads up Sebastian; this looks reasonable, but I'm sure the example code will need to be addressed.

darthtrevino (approved):

eps1lon:
@darthtrevino How do you currently type-check the repository? I see various `tsconfig` but couldn't find a single script that type-checks all releveant files.

darthtrevino:
It's a little convoluted because we run a pipeline of `Typescript` -> `Babel` with CJS and ESM output. We do this to gain the benefit of `@babel/preset-env` and using a browserlist query to determine what language features to incorporate, and what to use `core-js` for. In the future I'm hopeful that this can be simplified.
 
 Each TS project references the `build` project within the monorepo, which provides a common set of gulp tasks: 

https://github.com/react-dnd/react-dnd/blob/a30549039ec4b3a112982ddad14d8df7066f90b7/packages/dnd-core/gulpfile.js#L2

The custom build task here performs the orchestration: https://github.com/react-dnd/react-dnd/blob/main/packages/build/src/preset/defineBuild.ts

Issue #2621, opened 2020-06-19
Extreamly LOW performance on useDragLayer with a sortable list.


Kivylius:
**Describe the bug**
I been working with react-dnd for a few weeks, but I'm finding the library more and more frustrating everyday. I've used the example from sortable list + added a drag layer and the perfomance is extremely poor, even on my new MacBook Pro. The CPU is on 99%. I'm building my own library to create web apps like VisualCode or Photopea called [reactware](https://github.com/warejs/reactware) and dnd is quite important to this library, and maybe react-dnd is not the right tool for this job.

**Reproduction**

[Live Reproduction](https://codesandbox.io/s/hidden-framework-vqjys)

1. Go to codesandbox demo
2. Drag one of Panels Around.
3. Notice Lag

**Expected behavior**
A lag free experience, similar to what to what you get from the likes of jQuery or Interact.js

**Screenshots**

![ezgif-6-ee37fa524045](https://user-images.githubusercontent.com/1041335/85155995-5d8f7800-b251-11ea-95a7-64be860a82e7.gif)

<img width="1037" alt="Screenshot 2020-06-19 at 17 07 28" src="https://user-images.githubusercontent.com/1041335/85156415-f3c39e00-b251-11ea-9b08-ea34a06ab21a.png">

**Desktop (please complete the following information):**

- OS: MacOS 13.15.5
- Browser Chrome
- Version 83.0.4103.106

**Other**

If we can take a look at the example made in 2016 using react dnd decorators:

http://react-trello-board.web-pal.com/# 

We can see quite nice performance, but it was done using decorators and not hooks, maybe hooks not meant for react-dnd if you looking for performance?

Related issues: #2414 , #592

Merynek:
Some news? I have this issue as well.

jcampbell05:
I looked at the implementation of the HTML 5 backend. I was able to dramatically improve performance by wrapping the `hover` action call with a request animation frame and only sending the latest mouse coordinates.

What is happening is react-dnd is sending every single mouse move to redux in order to handle dragOver events so these can queue up. Since the main reason you consume these events is to show UI to the user we only care about them on repaints.

By only sending the latest value for requested animation frame we are reducing the work react needs to do
check out our patched version here 

See if this helps you at all https://github.com/PSPDFKit-labs/react-dnd

We still have some lagging but I believe thats due to the implementation of our components rather than the library

fromi:
[...]

PR #3313, opened 2021-10-07
Throttle actions.hover in HTML5Backend for better performance

martikaljuve:
Hey! 

This PR is basically a condensed version of the existing PR by @jcampbell05: https://github.com/react-dnd/react-dnd/pull/3118

We encountered the same issue with slow/laggy drag performance when implementing a table component with reorderable columns. This change boosted the performance to an acceptable level, though profiling with DevTools still shows React work taking >20ms per animation frame (in [useCollector.ts](https://github.com/react-dnd/react-dnd/blob/main/packages/react-dnd/src/hooks/useCollector.ts)) in our use case, so there is room for improvement. 

Relevant thread: https://github.com/react-dnd/react-dnd/issues/2621


istateside:
I just tested this change out and it was a massive, noticeable performance improvement for us. @darthtrevino any way we can push forward a review on this?

darthtrevino (approved):
👍- looks great; thanks for condensing the original PR!

darthtrevino:
It looks like this breaks the tests; I'll merge this once the tests are fixed and the build passes

istateside:
Thanks for the quick turnaround @darthtrevino ! And thanks @martikaljuve and @jcampbell05 for the original fix :+1:

istateside:
Hey @darthtrevino any chance we could get a new version published to NPM with the recent changes that have been merged?

darthtrevino:
Yeah I'll cut a version when I get home. One issue that will probably come up is that we'll need to put a conditional check around requestAnimationFrame so that it will work in SSR environments

aodinok:
We are facing issues in our tests with: `Invariant Violation: Expected targetIds to be registered.` error.
Just moving tests around and re-ordering them makes them randomly succeed/fail. 
I tried adding cleanup after each of the tests but problem is still there. It feels strange that this started to happen only after last version release and this comment in tests makes me wonder that there was race condition introduced:

[```
Invariant Violation: Expected targetIds to be registered.
```](https://github.com/react-dnd/react-dnd/pull/3313/commits/6ae98cb441c82986328ae9f2e95011f4ea1068c3#diff-0dfbd8f890969bccce471b9f6b210fc0856e2c1a0074c307ecba20ffd41f9841R103)

**Comment by beewolf233** [comment]
> 

HEELO do you success solve it?? I have same question too

File: packages/backend-html5/src/HTML5BackendImpl.ts [hunk 1]
```diff
@@ -601,10 +604,19 @@ export class HTML5BackendImpl implements Backend {
 		}
 
 		this.altKeyPressed = e.altKey
+		this.lastClientOffset = getEventClientOffset(e)
+
+		if (this.hoverRafId === null) {
+			this.hoverRafId = requestAnimationFrame(() => {
+				if (this.monitor.isDragging()) {
+					this.actions.hover(dragOverTargetIds || [], {
```
vsiao, line 612: Hi, we're experiencing the same issue as @aodinok here with

```
Invariant Violation: Expected targetIds to be registered.
```

and we're wondering whether this should be considered a bug within react-dnd. Since `requestAnimationFrame` is asynchronous, it seems problematic to capture `dragOverTargetIds` at time A and expect it to be accurate when the callback is executed at time B.

More philosophically, is it this library's responsibility to throttle these callbacks? From what I understand, React doesn't throttle `onScroll` callbacks either, so the argument could be made that this should really be implemented in application/user code instead.

Let me know if I should create a separate issue for this—thanks!

Issue #3193, opened 2021-05-04
Sortable Example Issue


wjgogogo:
**Describe the bug**

In the [Sortable Example](https://codesandbox.io/s/github/react-dnd/react-dnd/tree/gh-pages/examples_hooks_js/04-sortable/simple?from-embed=&file=/src/Container.jsx), if I drag the item and move it so fastly, the rest of items will be disorder.

**Reproduction**

https://user-images.githubusercontent.com/17964556/116967171-4e3be080-ace4-11eb-9b27-98c457049c4f.mov

For better demonstration, I replace the default text with "1,2,3,4,5", I only drag "5", so rests should keep the order like "1, 2, 3, 4",  but actually they are not.

And this scenario only happened with hooks, decorators work fine.

[Live Reproduction](https://codesandbox.io/s/github/react-dnd/react-dnd/tree/gh-pages/examples_hooks_js/04-sortable/simple?from-embed=&file=/src/Container.jsx)

**Desktop (please complete the following information):**

- Browser [e.g. chrome]
- Version [e.g. 90]

metafoo:
Stumbled upon the same issue using Safari 14.0.3

moeatsy:
have same issue using chrome 89.0

fsp:
Here's a _potential_ fix: React's dependency array in useCallback/useEffect wasn't designed with objects/array support in mind. At one point or another the `cards` state may become stale. My hunch is that it can be fixed using a Ref:


```
    const cardsRef = useRef(cards);

    const moveCard = useCallback((dragIndex, hoverIndex) => {
      const dragCard = cardsRef.current[dragIndex];
      cardsRef.current = update(cardsRef.current, {
        $splice: [
          [dragIndex, 1],
          [hoverIndex, 0, dragCard]
        ]
      });
      setCards(cardsRef.current);
    }, []);
```

[Demo](https://codesandbox.io/s/tender-meadow-ng22c?file=/src/Container.jsx:593-943)

mcrowder65:
This is also affecting my current project. Unfortunately using setCards with prevCards is not fixing it :\

bezsinnyidmytro:
Faced the same issue, but solved it in slightly different way than @SirKadogan propose.
```
const [orderedCards, setOrderedCards] = useState(cards);
const moveCard = useCallback((dragId, hoverIndex) => {
  const nextCards = orderedCards.slice();
[...]

PR #3314, opened 2021-10-07
Updating sortable example to use useState setter as function

SirKadogan:
This solves #3193 

Basically when moving the cards around, the state is not able to keep up with the movement. 

By using the inner state as a function, the problem is solved.


darthtrevino:
Hey 👋 - thanks for the PR, but I'm getting some "unrelated history" conflicts when trying to merge main. I'll cut a new PR

PR #3346, opened 2021-12-03
Expose original dataTransfer object

JacksonGariety:
The browser's `dataTransfer` object must be exposed during drop events.

This is because, per [this MDN article](https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API):

> A key difference between the DataTransfer and DataTransferItem interfaces is that the former uses the synchronous getData() method to access a drag item's data, but the latter instead uses the asynchronous getAsString() method.

And the **react-dnd `DropTarget` does not allow the `drop()` handler to be an `async function`**.

Therefore, in order to access data of a `drop` event in the handler, one must use the `getData` function which is only available on the original `DataTransfer` object. This PR exposes this object to the user.


darthtrevino (approved):
Thank you for the PR and the background details on why this was necessary

PR #3379, opened 2022-02-06
fix: revert a recent PR that made the draggable attribute conditional

### LINKED ISSUE #3378: Can't drag multiple times in 15.x release
**Describe the bug**
Since 15.x release it seems that I can no longer drag an item more than once. Only the first drag works. Subsequent drags are not enabled. Issue appears to be in react-dnd-html5-backend and the last working version was 14.1.0. I initially reported this issue in https://github.com/react-dnd/react-dnd/pull/3187 but I"m unsure if it's related to that and I think it definitely deserves it's own issue now.

**Reproduction**

You can see the issue in the chess example I have below on codesandbox. With 15.x of react-dnd-html5-backend dependency you can only move the chess piece once. If you change back to 14.1.0 you will be able to move the chess piece multiple times.

https://codesandbox.io/s/you-only-drag-once-9iszg

https://user-images.githubusercontent.com/8976377/152684323-2bb58547-b0cb-4d37-847f-d9ccd134313c.mov

**Expected behavior**
Able to drag and drop items more than once.

darthtrevino:
Fixes #3378

### DIFF
--- a/.yarn/versions/196dbc06.yml
+++ b/.yarn/versions/196dbc06.yml
@@ -1,0 +1,9 @@
+releases:
+  react-dnd-html5-backend: patch
+
+declined:
+  - react-dnd-documentation
+  - react-dnd-examples
+  - test-suite-cra
+  - test-suite-vite
+  - react-dnd-test-utils
--- a/.yarn/versions/196dbc06.yml
+++ b/.yarn/versions/196dbc06.yml
@@ -1,9 +1,9 @@
 releases:
   react-dnd-html5-backend: patch
+  react-dnd-examples: patch
 
 declined:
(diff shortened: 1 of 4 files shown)


dmbarry86:
Is this included in 15.1.2 release? I don't see any mention of it in release notes but notice it has been merged so just wanted to check.

darthtrevino:
Yup, it was in 15.1.1

dmbarry86:
> Yup, it was in 15.1.1

Thanks. Incidentally I don't see a 15.1.1 release. Just seems to jump from 15.1.0 to 15.1.2. Anyway thanks for the fix.

Issue #3383, opened 2022-02-07
Named export 'asap' not found / NextJS


varunarora:
**Describe the bug**
Hi! I am a DnD newbie, and have been out of all of the details of the JavaScript/webpack moduling and tooling world for a while, so can't figure this out even though this might be really easy for some of you to figure out.

I am using NextJS (12.0.10) and React (17.0.2). I don't access to their internal webpack config. When trying to import the DnD `useDrag` export (version 15.1.0) for the first time, I see this error:

```
SyntaxError: Named export 'asap' not found. The requested module '@react-dnd/asap' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from '@react-dnd/asap';
const { asap } = pkg;
```

Begins with this call stack item:
```
dnd-core/dist/esm/classes/HandlerRegistryImpl.mjs (6)
```

**Reproduction**

1. `npx create-react-app`
2. cd `<project_name>`
3. `npm install --save react-dnd react-dnd-html5-backend`
4. In `pages/index.js`, import using `import { useDrag } from 'react-dnd'`

**Additional context**
A downgrade to v14 makes the issue go away.

darthtrevino:
This is definitely something that's fixable, and I'll plan on verifying that those dependencies are ESM. 

I also added a Next.JS tester in the app here (https://github.com/react-dnd/react-dnd/pull/3384), and it appears to work fine as-is

darthtrevino:
I've updated the asap dependency to export ESM; let me know if this issue continues

varunarora:
Wow thank you @darthtrevino you are so quick and amazing!

dustin-H:
Hi @darthtrevino,
thanks for fixing that issue.

It seems to be released in version `15.1.2`. (https://github.com/react-dnd/react-dnd/releases/tag/v15.1.2)
However, this versions hasn't been released on npm.

Could you please release it there? 😃 
Thanks! 🍻

PR #3384, opened 2022-02-07
Add Next.JS App Tester Package

darthtrevino:


packages/dnd-core/package.json
```diff
@@ -24,8 +24,8 @@
     "url": "https://github.com/react-dnd/react-dnd.git"
   },
   "dependencies": {
-    "@react-dnd/asap": "^4.0.0",
-    "@react-dnd/invariant": "^2.0.0",
+    "@react-dnd/asap": "workspace:packages/util-asap",
```
kewitz, line 27: @darthtrevino I don't think this will work for your users when they install `react-dnd` from NPM.

From my package-lock:
```
"@react-dnd/asap": {
      "version": "4.0.0",
      "resolved": "https://registry.npmjs.org/@react-dnd/asap/-/asap-4.0.0.tgz",
      "integrity": "sha512-0XhqJSc6pPoNnf8DhdsPHtUhRzZALVzYMTzRwV4VI6DJNJ/5xxfL9OQUwb8IH5/2x7lSf7nAZrnzUD+16VyOVQ=="
    },
```

darthtrevino: I moved the asap fork into this monorepo. The workspace: links will link packages together at dev time, but at publication time yarn will swap out workspace: links for version numbers. 

Are you having issues with the latest version resolving the @react-dnd/asap packages?

kewitz: I'm not completely sure, but I don't think this fixes https://github.com/react-dnd/react-dnd/issues/3383 outside the dev environment.
Perhaps the safe bet is to just import the default export and deconstruct as the error suggests?
```js
import pkg from '@react-dnd/asap';
const { asap } = pkg;
```

PR #3392, opened 2022-02-15
chore: remove file extensions from imports

amorriscode:
### tl;dr

- [x] Removes file extension (`.js`) from component imports in examples

### Description

I was taking a look at [a TypeScript example on CodeSandbox](https://codesandbox.io/s/github/react-dnd/react-dnd/tree/gh-pages/examples_ts/02-drag-around/custom-drag-layer) and saw that it didn't compile. The reason was that the imports were looking for `*.js` files, even though they don't exist. This PR removes the file extensions so the components can be loaded properly.

### Resources

- [Official CodeSandbox example that fails to compile](https://codesandbox.io/s/github/react-dnd/react-dnd/tree/gh-pages/examples_ts/02-drag-around/custom-drag-layer)


darthtrevino:
The extensions are for ESM compatibility. We can dig into fixing issues with the examples, but the module tests need to be able to pass

amorriscode:
Ah yeah, sorry about that @darthtrevino. I should've done a bit more research. 😞 I'll see if I can find a solution over the next few days. If not, I'll close out the PR.

darthtrevino:
I've thought about this some more, and I think it's probably safe to remove the extensions in the examples/ package only. My main concern was with the core libraries, but we use transpiled versions of those, and support for TS imports like this are still emerging in common toolchains.

Issue #2722, opened 2020-08-12
Website example code has exception


hiyangguo:
**Describe the bug**
![image](https://user-images.githubusercontent.com/15609339/90006078-cf70c480-dcca-11ea-8299-5fbc922acae8.png)

**Reproduction**

[Live Reproduction](codesandbox.io/....)

1. Go to https://react-dnd.github.io/react-dnd/examples/sortable/cancel-on-drop-outside
2. Click on `Using Hooks` button.
3. Switch to `Typescript` tab.

**Expected behavior**

**Screenshots**
![image](https://user-images.githubusercontent.com/15609339/90006179-0050f980-dccb-11ea-9679-6253294c7b9b.png)

**Desktop (please complete the following information):**

- OS: [e.g. iOS] MAC
- Browser [e.g. chrome, safari] chrome 
- Version [e.g. 22]  83

mburakeker:
Desktop:
OS: Windows
Browser: Chrome
Version: 85

And can confirm this.

PR #3395, opened 2022-02-20
docs(#2722): remove .js extension from example inputs

KevinKoetz:
Removed the .js extensions from Imports to resolve the error shown when navigating to the Typescript examples on CodeSandbox.
See: https://github.com/react-dnd/react-dnd/issues/2722

I tested the changes by building the examples and the Docsite and checked in Codesandbox if removing the .js extension solves the issue and keeps the JS Version working.


KevinKoetz:
Missed #3392 and your comment about removing the extensions.

darthtrevino:
I ended up removing the extensions from the examples - support for TS using .js extension imports is still a WIP in most bundlers.

Issue #3403, opened 2022-03-08
Uncaught Invariant Violation: Expected targetids to be registered after dragging too fast


takeandrei5:
Hello!

We implemented a two column list using the react hooks and hover functionality. The functionality works great and there are no issues if we move the items inside the same list (reordering) or if we drag the items between the two lists in a normal manner. However, when we try to drag the items between the two lists as fast as we can, we randomly get this error. We looked into all the other topics which we could find but to no avail. 

Any ideas?

- OS: Windows
- Browser Chrome 
- Version 98.0.4758.102 

We are running the latest version of react-dnd. v15.11

RobinKamps:
Same here. Unfortunately after the error occures the drag preview gets very laggy, so its hard to just ignore the error. Seems to be the same as #763.

Youmyin:
I tried to change to the latest version 15.1.1, but it still doesn't work. Surprisingly, this error only exists in the local development environment and disappears after deploying to the server, so how can i fixed it?

PR #3409, opened 2022-03-22
fix: "Invariant Violation: Expected targetIds to be registered."

### LINKED ISSUE #763: Invariant Violation: Expected targetIds to be registered.
Hi there. We are getting this error (although the functionality is okay and works fine). Do you have any idea how it can happen? couldn't figure it out from the source code.

joshkel:
We've sporadically seen this error from customers. The call stack indicates that it's originating from HTML5BackendImpl's
handleTopDragOver's requestAnimationFrame callback.  I've been unable to reproduce it locally; however, if I simulate a slowdown by replacing `requestAnimationFrame(callback)` with `setTimeout(callback, 10000)`, I can fairly reliably reproduce this error.

To fix it, I believe HTML5BackendImpl should consistently clear the hover animation whenever the drag operation is ended for any reason.

I locally tested this fix in the v15.1.2 tag, and it appeared to work.

Fixes #763, #3403

### DIFF
--- a/packages/backend-html5/src/HTML5BackendImpl.ts
+++ b/packages/backend-html5/src/HTML5BackendImpl.ts
@@ -406,6 +406,9 @@ export class HTML5BackendImpl implements Backend {
 			this.hoverRafId === null &&
 			typeof requestAnimationFrame !== 'undefined'
 		) {
+			// cancel any existing hover if present
+			this.cancelHover()
+
 			this.hoverRafId = requestAnimationFrame(() => {
 				if (this.monitor.isDragging()) {
 					this.actions.hover(dragOverTargetIds || [], {
--- a/packages/backend-html5/src/HTML5BackendImpl.ts
+++ b/packages/backend-html5/src/HTML5BackendImpl.ts
@@ -347,6 +347,7 @@ export class HTML5BackendImpl implements Backend {
 		if (this.clearCurrentDragSourceNode() && this.monitor.isDragging()) {
 			this.actions.endDrag()
 		}
+		this.cancelHover()
 	}
 
 	private setCurrentDragSourceNode(node: Element | null) {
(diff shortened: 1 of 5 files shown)


darthtrevino (approved):

packages/backend-html5/src/HTML5BackendImpl.ts
```diff
@@ -400,6 +401,33 @@ export class HTML5BackendImpl implements Backend {
...
 	}
 
+	private scheduleHover = (dragOverTargetIds: string[] | null) => {
+		if (
+			this.hoverRafId === null &&
+			typeof requestAnimationFrame !== 'undefined'
+		) {
+			this.hoverRafId = requestAnimationFrame(() => {
+				if (this.monitor.isDragging()) {
+					this.actions.hover(dragOverTargetIds || [], {
```
vsiao, line 411: Thanks for trying to fix this bug! I just commented on https://github.com/react-dnd/react-dnd/pull/3313/files#r836932854 and I think it's still relevant here so I'm cross-linking it. From what I can tell, it still seems very possible to encounter bad race conditions, even with the changes that you've made.

Did you try always calling `cancelHover` before scheduling to ensure that we're capturing the most recent copy of `dragOverTargetIds`?

packages/backend-html5/src/HTML5BackendImpl.ts
```diff
@@ -406,6 +406,9 @@ export class HTML5BackendImpl implements Backend {
 			this.hoverRafId === null &&
 			typeof requestAnimationFrame !== 'undefined'
 		) {
+			// cancel any existing hover if present
+			this.cancelHover()
```
vsiao, line 410: @darthtrevino this fix that you added is always a no-op since it's inside the `if` condition—it should be moved before the `if`.

darthtrevino: Gotcha, thanks - is this even necessary then?

packages/backend-html5/src/HTML5BackendImpl.ts
joshkel: Thanks for the feedback, @vsiao. So, if I'm understanding correctly:

In the original code, if the user hovers over element A, then quickly hovers over element B before the requestAnimationFrame for A can fire, then the hover will be out of date.

After my changes, I _think_ this won't throw errors, because I think the code is structured to cancel any hover effects before anything that would invalidate a target ID. (Actually, it may be possible for an externally triggered rerender to invalidate a target ID. That may need more investigation.) But it's still a visual bug.

To address that, you're saying we should also cancel the hover effect before triggering any hover effect?

That seems correct to me (and so @darthtrevino's [fix](https://github.com/react-dnd/react-dnd/pull/3409#pullrequestreview-923919531) should go outside of the `if`, like you said).

joshkel:
@omridevk, if you can share a test project or a [codesandbox.io](https://codesandbox.io) demo, I can investigate further.

omridevk:
i'll try to, it just a private project that I am working on, I'll try to create a minimal reproduction example.

RobinKamps:
Unfortunately this error still persists - it occurs on nested lists, most times if you drag around like crazy.
Here is an angular :-) example:

Demo:
https://stackblitz.com/github/RobinKamps/dnd-test

Repo:
https://github.com/RobinKamps/dnd-test

miracle0930:
Not sure if this is a correct fix. Currently, ```monitor.isOver()``` may be ```false``` even you have already hovered into it. Seems that some ```requestAnimationFrame``` is not canceled then the ```scheduleHover``` mistakenly dispatches a hover action with empty ```targetIds``` to the reducer. I tried and found that adding one more ```cancelOver``` at the end of ```handleTopDragEnter``` will resolve this issue. 
image

PR #3426, opened 2022-04-08
Fix single-target-in-iframe/src/Example.js (`.js` instead of `.jsx`)

meselgroth:
Example is pointing to `Container.js` file instead of `Container.jsx`

There are other imports/exports with incorrect extensions. If this is useful, I can fix those too.


darthtrevino:
Thanks for bringing this up; I think the right solution here is to rename the .jsx files in the dist/docs folder of the examples project to .js. I've opened a PR to add this build step here: https://github.com/react-dnd/react-dnd/pull/3427

PR #3432, opened 2022-04-13
Fix for "Invariant Violation: Expected targetIds to be registered."

marikobi:
We keep seeing this error during drag and drop, and I believe it's happening because we give the list of targetIds to the async hover function and expect that list to be correct when we actually execute it : see https://github.com/react-dnd/react-dnd/pull/3313#pullrequestreview-923862465

Moving the invariant check after removing non-matching Ids should fix it since if the targetId doesn't exist any more it'll be removed from the targetIds list.

Tested locally by patching this change on V14.0.3 and this fix seems to be working.

Fixes https://github.com/react-dnd/react-dnd/issues/763, https://github.com/react-dnd/react-dnd/issues/3403


darthtrevino (approved):
This looks good to me; thanks for the PR!

marikobi:
hi @darthtrevino, thanks for approving this fix. Since it seems like other people are hitting the same issue, would you consider releasing the version 14 with this patch?

darthtrevino:
@marikobi  I don't usually maintain old major releases of this library (just for my own sanity's sake) - is there anything preventing you from using v16?

marikobi:
>@marikobi I don't usually maintain old major releases of this library (just for my own sanity's sake) - is there anything preventing you from using v16?

We are using Decorators API in our codebase so would need to first migrate to using hooks before we start using versions above 14

PR #3437, opened 2022-04-19
Fix Typescript examples to allow for compilation

SeenicRoot:
Mostly fixing imports to match Typescript modules style.


darthtrevino:
Thanks for bringing this issue to attention. I'm going to go with a post-processing step instead so we can preserve the ESM style imports in the example source: https://github.com/react-dnd/react-dnd/pull/3439

PR #3484, opened 2022-08-09
Fix typings for `monitor.getItem()`

akphi:
I have noticed that this method should be annotated as potentially returning `null`. What do you think @darthtrevino ?


akphi:
@darthtrevino is there anything else I could do here to get this fix in?

jtbandes:
Would be a great fix! This seems like a no-brainer :)

pieralukasz (approved):

Michi-2142 (approved):

PR #3498, opened 2022-09-06
feat: allow configuring when to use the 'copy' dropEffect

### LINKED ISSUE #2979: Shift instead of Alt to copy?
Is it possible to use shift modifier to copy items instead of moving them? Every app uses shift for this purpose. Why is it hardcoded to Alt instead?

        return {
            dropEffect: this.altKeyPressed ? 'copy' : 'move',
            ...(sourceNodeOptions || {}),
        };

TheUnlocked:
Allows using keys other than alt to set the dropEffect to 'copy', and allows using a predicate callback in case the user wants more complex behavior.

Resolves #2979

### DIFF
--- a/packages/backend-html5/src/OptionsReader.ts
+++ b/packages/backend-html5/src/OptionsReader.ts
@@ -35,4 +35,21 @@ export class OptionsReader {
 	public get rootElement(): Node | undefined {
 		return this.optionsArgs?.rootElement || this.window
 	}
+
+	public isCopying(e: DragEvent): boolean {
+		if (this.optionsArgs?.copyKey === undefined) {
+			return false
+		}
+		switch (this.optionsArgs?.copyKey) {
+			case 'shift':
+				return e.shiftKey
+			case 'ctrl':
+				return e.ctrlKey
+			case 'alt':
+				return e.altKey
+			case 'meta':
+				return e.metaKey
+		}
+		return this.optionsArgs?.copyKey(e)
+	}
 }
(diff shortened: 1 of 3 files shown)


packages/backend-html5/src/OptionsReader.ts
```diff
@@ -35,4 +35,21 @@ export class OptionsReader {
...
+		}
+		switch (this.optionsArgs?.copyKey) {
+			case 'shift':
+				return e.shiftKey
+			case 'ctrl':
+				return e.ctrlKey
+			case 'alt':
+				return e.altKey
+			case 'meta':
+				return e.metaKey
```
vlad00777, line 51:
```suggestion
		switch (this.optionsArgs?.copyKey) {
			case 'shift':
				return e.shiftKey
			case 'ctrl':
				return e.ctrlKey
			case undefined:
			case 'alt':
				return e.altKey
			case 'meta':
				return e.metaKey
```

vlad00777 (changes requested):

packages/backend-html5/src/OptionsReader.ts
TheUnlocked: It looks like this won't type check without a non-null assertion which is forbidden by the linter. Seems to be a TypeScript issue (minimal reproduction: https://www.typescriptlang.org/play?ts=4.9.0-dev.20221007#code/CYUwxgNghgTiAEYD2A7AzgF3gDwFzwG94AjWAfgAoBKfANyQEth4BfeAH3gFcVQAzBihDAA3ACg+PMBgap4fJEmqEx8NfDQB3BhjAALeBWxkAdKRhUV664ihoEPfoOG5VN93AxcYKcdZZuathmsNTiAUA).

There was an unnecessary optional chaining operator there though which I've removed.

Issue #3529, opened 2022-12-30
How to change drag dropEffect based on drop targets?


sdt-jchau:
The current dropEffect example shows how to change dropEffect on the drag source based on the drag source itself (https://react-dnd.github.io/react-dnd/examples/customize/drop-effects).  Is there a way to dynamically change the dropEffect based on the target?  For example, I have 2 drop target boxes.  If I drag an item to Box 1, it should show Add effect.  If I drag an item to Box 2, it should show Copy effect.

josephdangerstewart:
It doesn't look like this is supported out of the box, but I was able to get something working with this hack (loosely inspired by [this hack](https://github.com/react-dnd/react-dnd/issues/512#issuecomment-724709481))

```ts
export const HTML5BackendWithModifierKeys = (manager, context) => {
	const result = HTML5Backend(manager, context);
	result.dropTargetOptions = {};

	result.handleTopDragOver_ = result.handleTopDragOver;
	result.handleTopDragOver = (e) => {
		const { dragOverTargetIds } = result;
		const bestDropTarget = (dragOverTargetIds ?? []).find(
			(x) => result.monitor.canDropOnTarget(x) && x
		);

		const dropTargetDropEffect =
			bestDropTarget &&
			result.dropTargetOptions[bestDropTarget]?.dropEffect;

		result.handleTopDragOver_(e);
		result.altKeyPressed = Boolean(dropTargetDropEffect);
		e.dataTransfer.dropEffect =
			dropTargetDropEffect || e.dataTransfer.dropEffect;
	};

	result.connectDropTarget_ = result.connectDropTarget;
	result.connectDropTarget = (targetId, node, options) => {
		result.dropTargetOptions[targetId] = options;
		const unsubscribe = result.connectDropTarget_(targetId, node, options);

		return () => {
			delete result.dropTargetOptions[targetId];
			unsubscribe();
		};
	};

	return result;
};
```

Then you can specify the `dropEffect` in an options object in the `useDrop` hook like so

```js
useDrop({
	...,
	options: {
		dropEffect: 'copy'
	},
})
```

PR #3531, opened 2023-01-07
feat: support setting drop effect for drop targets.

josephdangerstewart:
Adds an option to the `useDrop` hook to set the drop effect based on drop target and updates the documentation to reflect the changes.

Since there can be multiple sources of truth for what the drop effect should be (drag source options, drop target options, and alt key being pressed), my PR uses the following precedence (but I'm happy to change it if the reviewer prefers something different)

1. Use dropEffect from drag source
2. Use dropEffect from the most deeply nested drop target
3. If the alt key is pressed, then use "copy"
4. Default to "move"

Resolves: #3529


cliffhall:
Anybody still interested in making this work?

PR #3565, opened 2023-04-27
fix: don't lose drag source/preview if disconnected and immediately reconnected

jtbandes:
This fixes a bug demonstrated in the following codesandbox. One drag source works, and the other does not.
https://codesandbox.io/s/xenodochial-christian-nu3rlw?file=/src/App.tsx

https://user-images.githubusercontent.com/14237/234934158-9208f10a-172c-4470-93aa-9cb6539b8ff2.mov

I first encountered this bug in 14.0.3, so I think it was introduced in https://github.com/react-dnd/react-dnd/pull/3290. At the time I [skipped upgrading this package](https://github.com/foxglove/studio/pull/1849) because I couldn't resolve it. Here we are 1.5 years later and it seems like the bug is still present 😅

The bug is triggered by the combination of `useDrag({ ... options: { dropEffect: "copy" } })` and `useEffect(() => setValue(1))`:
- The `setValue` call triggers the component to render again.
- A new options object `{ dropEffect: "copy" }` is allocated and passed into useDrag.
- Since `dragSourceOptions` is a dependency for this `useLayoutEffect`, `connector.disconnectDragSource()` and `connector.reconnect()` are called: https://github.com/react-dnd/react-dnd/blob/7c88c37489a53b5ac98699c46a506a8e085f1c03/packages/react-dnd/src/hooks/useDrag/useDragSourceConnector.ts#L20-L24
- Inside `reconnectDragSource()`, didChange is set to false because didDragSourceOptionsChange returns false. It returns false because the new `dragSourceOptions` are `shallowEqual` to the previous options. https://github.com/react-dnd/react-dnd/blob/7c88c37489a53b5ac98699c46a506a8e085f1c03/packages/react-dnd/src/internals/SourceConnector.ts#L101-L107 https://github.com/react-dnd/react-dnd/blob/7c88c37489a53b5ac98699c46a506a8e085f1c03/packages/react-dnd/src/internals/SourceConnector.ts#L179-L184

This patch fixes the bug by adding extra state variables, `isDragSourceConnected` & `isDragPreviewConnected`. When these are false, `backend.connectDragSource` & `backend.connectDragPreview` will be called.

### DIFF
--- a/.yarn/versions/42579e0f.yml
+++ b/.yarn/versions/42579e0f.yml
@@ -1,0 +1,9 @@
+releases:
+  react-dnd: patch
+
+declined:
+  - react-dnd-documentation
+  - react-dnd-examples
+  - test-suite-cra
+  - test-suite-vite
+  - react-dnd-test-utils
--- a/packages/react-dnd/src/internals/SourceConnector.ts
+++ b/packages/react-dnd/src/internals/SourceConnector.ts
@@ -53,6 +53,8 @@ export class SourceConnector implements Connector {
 	private dragPreviewOptionsInternal: DragPreviewOptions | null = null
 	private dragPreviewUnsubscribe: Unsubscribe | undefined
 
+	private isDragSourceConnected = false
+	private isDragPreviewConnected = false
 	private lastConnectedHandlerId: Identifier | null = null
 	private lastConnectedDragSource: any = null
 	private lastConnectedDragSourceOptions: any = null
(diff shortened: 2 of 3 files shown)


jtbandes:
Hi @darthtrevino – do you think there's a chance you'd have time to look at this in the near future, or do you know when you might get a chance? Thanks for your work on this library 🙌

jtbandes:
Hi @pieralukasz, wondering if you would be able to take a look at this, in case you are working on this project nowadays. Currently I'm still maintaining a patch for it in my own app :)

PR #3659, opened 2025-03-27
Fix missing .js import in connectors.ts

### LINKED ISSUE #3658: Missing .js extension in import cause typescript errors
**Describe the bug**
There is an import without js extension, causing TypeScript error when using react-dnd

**Reproduction**

See https://github.com/react-dnd/react-dnd/blob/1dd1bcdd39d9545fdbb319e34ddbeecf0fb9d89c/packages/react-dnd/src/types/connectors.ts#L3

**Expected behavior**
.js is added

binhpv:
Fix #3658

### DIFF
--- a/packages/react-dnd/src/types/connectors.ts
+++ b/packages/react-dnd/src/types/connectors.ts
@@ -1,6 +1,6 @@
 import type { ReactElement, RefObject } from 'react'
 
-import type { DragPreviewOptions, DragSourceOptions } from './options'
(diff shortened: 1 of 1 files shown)


binhpv:
Hi @darthtrevino , during our ESM conversion we found this issue, and have to skipLibCheck to overcome TypeScript error.
Would be great to have this merged and released, if possible.

Many thanks!

PR #3664, opened 2025-05-24
TouchBackend: Clear old timeout in handleTopMoveStartDelay

raineorshine:
## Problem

Normally, `delayTouchStart` requires that the user tap and hold the draggable element for a given delay before a drag is initiated. If there is a touchmove during this time, the drag is cancelled.

However, if the user quickly touches the screen and then initiates a tap + move within `delayTouchStart`, **the drag will incorrectly start**, ignoring the move that would have cancelled it.

A real-world case is described in https://github.com/cybersemics/em/issues/2967.

## Cause

This occurs because the `handleTopMoveStart` timer is only cleared in [`handleTopMove`](https://github.com/react-dnd/react-dnd/blob/1dd1bcdd39d9545fdbb319e34ddbeecf0fb9d89c/packages/backend-touch/src/TouchBackendImpl.ts#L416). When `handleTouchStartDelay` is triggered in quick session (by two taps), then the second call to `handleTouchStartDelay` will overwrite the existing timer before `handleTopMove` has the chance to clear it. Thus, when `handleTopMove` fires, it only clears the second timer, and the first incorrectly resolves `handleTouchStart` thinking that the user has not moved their finger.

## Solution

This is fixed by clearing the old timer before setting the new timer.

### DIFF
--- a/.yarn/versions/efe692e7.yml
+++ b/.yarn/versions/efe692e7.yml
@@ -1,0 +1,5 @@
+releases:
+  react-dnd-touch-backend: patch
+
+declined:
+  - react-dnd-documentation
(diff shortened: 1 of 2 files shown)


raineorshine:
@darthtrevino Ready for review! Thanks.

PR #3669, opened 2025-07-03
fix: Fix import which does not follow ESM rules

martinnaj:


martinnaj:
@pieralukasz Review please

pieralukasz (approved):

martinnaj:
@darthtrevino Please try to publish this change asap once you are back from vacation

